PR #14, opened 2017-02-03
Verify that the temp directory exists before runner

0v3lh4:
Error in windows when there is no c:/temp directory

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -138,6 +138,10 @@ func (r *Runner) Start(ctxt context.Context) error {
 	// set user data dir, if not provided
 	_, ok = r.opts["user-data-dir"]
 	if !ok {
+		if _, err = os.Stat(DefaultUserDataTmpDir); os.IsNotExist(err) {
+			os.Mkdir(DefaultUserDataTmpDir, os.ModeDir)
+		}
+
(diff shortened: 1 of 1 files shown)


kenshaw:
I appreciate the pull request, but I am not going to add this to the code, my rationale is below:

I do not believe that chromedp/runner should be creating/deleting directories. I understand that `c:\temp` is not available on Windows by default; however Windows does not in general have a standard temp directory path that I am aware of. I chose `c:\temp` simply so that it had some kind of parallels to Unix. The other issue with Windows is that on Unix systems, there is an expectation that `/tmp` is erased or is otherwise transient for the container/vm. I don't believe that behavior is the same with the various Windows temporary paths.

I definitely would like to come up with behavior that is "right" for Windows, and is also in line with what I think `chromedp` should be doing, but I believe that leaving the management of temporary file system resources should be left to the operating system, or barring that, should be managed by the user. If you have a different way that this could be accomplished (that doesn't require creating/deleting directories), I would greatly appreciate it!

0v3lh4:
In Windows 7+ there is the environment variable for a temporary directory of the user.
You can access it with os.Getenv("temp") or os.Getenv("tmp").

PR #18, opened 2017-02-12
Added new default logger that can be overridden from outside the chromedp package

JalfResi:
Found it difficult to separate my programs logs from the logs generated by chromedp, so this feature allows me to redirect or ignore chromedp logging

### DIFF
--- a/chromedp.go
+++ b/chromedp.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"log"
 	"sync"
 	"time"
 
@@ -110,14 +109,14 @@ func (c *CDP) AddTarget(ctxt context.Context, t client.Target) {
 	// create target manager
 	h, err := NewTargetHandler(t)
 	if err != nil {
-		log.Printf("error: could not create handler for %s, got: %v", t, err)
+		Logger.Printf("error: could not create handler for %s, got: %v", t, err)
 		return
 	}
 
 	// run
 	err = h.Run(ctxt)
 	if err != nil {
-		log.Printf("error: could not start handler for %s, got: %v", t, err)
+		Logger.Printf("error: could not start handler for %s, got: %v", t, err)
 		return
 	}
 
--- a/handler.go
+++ b/handler.go
@@ -154,7 +154,7 @@ func (h *TargetHandler) run(ctxt context.Context) {
 					h.qres <- msg
 
 				default:
-					log.Printf("ignoring malformed incoming message (missing id or method): %#v", msg)
+					Logger.Printf("ignoring malformed incoming message (missing id or method): %#v", msg)
 				}
 
 			case <-h.detached:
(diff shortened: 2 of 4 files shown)


JalfResi:
Hmm.. not sure how to write. unit test for this! Will investigate...

kenshaw:
This failed not because of your code, but because Travis had some kind of intermittent network issue. Anyway, this isn't going to be merged, because the approach for specifying a logger needs to be more flexible than this. I had intended, previously, to add a `logger` member to the various types that could then be passed on creation, or otherwise copied in, I just haven't gotten to that yet. Funny enough, however, I was looking at implementing it over the last weekend but instead was working on building out better Docker images.

JalfResi:
Not a problem. I can implement it that way if you prefer. This was just a stop gap for myself.

kenshaw:
Oh absolutely! the log.Printf statements were just stop gaps themselves! I really do appreciate the pull request, and if I did not have something else immediately planned (ie, will be available in the next day or two) would definitely have added this to master. As it is, I'm working on a bunch of interrelated things at the moment (such as fixing context timeouts, proper logging options, etc) that should provide the flexibility needed for configuring these things. The primary issue is that I'm trying to keep the API high level, but flexible enough to be able to provide options for these things in a way that's maintainable moving forward.

JalfResi:
May I make a suggestion that the logger passed to the constructors be an interface that adheres to the stdlib logger? Many third-party loggers are designed as drop-in replacements for sodlib logger and it can be nice to use those sometimes e.g. marking chromedp log messages in a different colour.

I didn't do that on my pull request and now I look at it I think that would have been a much more useful addition.

kenshaw:
Yes, that is the plan. However, I am likely to provide some kind of structured interface/option as well, as there is a lot of junk / output from the protocol that is not likely relevant.

Typically, for other packages I've written, I usually just have a 'log' func setting, whose type is just of `func(string, ...interface{})`, which is of course compatible with the various standard log types, as well as provides easy ways to hook your own logging system into it. If you had more concrete ideas, I'm definitely open to them! I have been caught up with other issues today, otherwise I would have already pushed something out.

JalfResi:
Excellent stuff! Look forward to it!

kenshaw:
If you're up for it, I'll put the new API up in just a bit (next couple hours), and if you could test it out / provide feedback on how it works, I'd appreciate it.

kenshaw:
I just pushed my revisions that contain a simple attempt at providing "cleaner" logging options. Please take a look, and let me know. Please note though, that I realized there are some design mistakes/issues with the API based on exposing / setting the logging, which is part of the work I'm doing concurrently with fixing the relationship between the "pool" and the CDP object, so this will likely change in the future, but this is the style of the API that I would imagine is most suited to this.

JalfResi:
Looks spot on! I'll give it a bash this evening

PR #20, opened 2017-02-14
Changes on Clear mechanism

ranch:
This changes fix the issue when clearing value on textarea element.


kenshaw:
Fixed the behavior issue by using the #text child of the textarea instead.

PR #23, opened 2017-02-15
Pool and CDP options to specify target timeout

JalfResi:
I was getting a lot of "timeout waiting for response from target" errors and noticed that the DefaultNewTargetTimeout was hard-coded. So I added a CDP Option and matching PoolOption to specify these when configuring a pool.

### DIFF
--- a/chromedp.go
+++ b/chromedp.go
@@ -34,6 +34,9 @@ type CDP struct {
 	// handlerMap is the map of target IDs to its active handler.
 	handlerMap map[string]int
 
+	// how long to wait for the target before timing out
+	newTargetTimeout time.Duration
+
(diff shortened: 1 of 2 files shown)


JalfResi:
I'll add tests for these

kenshaw:
Hey -- seriously, I do appreciate your submission, but this is another one of those things that the existing code (all the timeout stuff) was only put in temporarily, and will be moved to setting a value on the context. I actually have most of that written on a separate branch, but haven't pushed the changes, as I have not yet written up the documentation on how to use it. Please keep a look for the revisions over the next couple days.

JalfResi:
Cool no probs

PR #45, opened 2017-03-27
Added start maximized option

crackcomm:
Adds `--start-maximized` flag to chrome.

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -391,6 +391,11 @@ func WindowSize(width, height int) CommandLineOption {
 	return Flag("window-size", fmt.Sprintf("%d,%d", width, height))
 }
 
+// StartMaximized is the Chrome command line option to start with maximized window size.
+func StartMaximized() CommandLineOption {
+	return Flag("start-maximized", true)
+}
+
 // UserAgent is the Chrome command line option to set the default User-Agent
 // header.
 func UserAgent(userAgent string) CommandLineOption {


kenshaw:
I appreciate the work done, but it's not really feasible to add a `runner.CommandLineOption` for every single flag. The `CommandLineOption` type is simple enough that any using package can easily/quickly add their own flag set. I don't believe that `--start-maximized` would be as widely used/needed as the options for things like `--remote-debugging-port`, `--proxy-server`, etc. 

Eventually, a line in the sand needs to be drawn, considering the dozens (hundreds?) of command line options available to launch Chrome with -- and for the record, I'm drawing it here.

kenshaw:
I realized that my wording above might be taken as more "harsh" than intended. As such, I'd like to clarify that to add a `--flag` `runner.Flag` equivalent, I feel a flag needs to meet the following criteria:

1) Wide applicability in general
2) Meaningful to both regular Chrome and to Chrome's headless variants (ie, `headless_shell`)

While a `runner.StartMaximized` flag criteria meets 1, I don't believe that it passes 2. 

The expected use cases of `chromedp` is to drive Chrome in some kind of automated fashion, either for unit testing, scraping, profiling, or similar. As such, whatever using package is likely to be running Chrome in some kind of headless context, and will most likely be using fixed window sizes (ie, `runner.WindowSize`).  It's also worth noting that when running a browser in a headless context, there is only a virtual window size, and "maximized" likely has no relevant context or meaning. As such, I don't believe that starting a window "maximized" is a great enough use case to warrant its own `runner.Flag`, especially since the API supports passing generic flags with `runner.Flag`.

crackcomm:
@kenshaw it totally makes sense. I appreciate your explanation.

PR #47, opened 2017-03-28
Performance fix #46

### LINKED ISSUE #46: Performance problems
Tends to eat 400% CPU

Probably caused by https://github.com/knq/chromedp/blob/8eb44961e2c3fbb7b3b302184114c2eb2059c4b8/handler.go#L397

[Flame graph](http://ipfs.io/ipfs/Qme5e7xeEix5hYR1LhzLjihBV3xYWBegb7N22QEA2d1bY7)

crackcomm:
Flame graph after a fix: https://ipfs.io/ipfs/QmQ4C8xPaGyAGkwamr3a9uJX7AfDNE9uE1XrtgoRTBb6kh

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -396,8 +396,7 @@ func (h *TargetHandler) next() int64 {
 // GetRoot returns the current top level frame's root document node.
 func (h *TargetHandler) GetRoot(ctxt context.Context) (*cdp.Node, error) {
 	// TODO: fix this
-	ctxt, cancel := context.WithTimeout(ctxt, 10*time.Second)
-	defer cancel()
+	timeout := time.After(10 * time.Second)
 
 	var root *cdp.Node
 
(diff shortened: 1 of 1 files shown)


kenshaw:
I appreciate the work here, but this needs to be fixed in a different location, and is tied into the setting of overall timeouts on the selector settings. I suppose this is happening because likely whatever you're doing, you're causing Chrome to change a bunch of nodes, which invalidates the frame frequently, and causes the selectors to `time out` as it were. To be honest, this code needs to be overhauled in general, and moved into some kind of channel based approach with maximum timeouts set properly via context that don't cause the context to be created new each time. I had never liked the way that I was doing this originally, but it had been a quick, stop-gap "good enough" for what it is being used.

crackcomm:
This pull request fixes performance problem, it is just a better way to handle timeout.
I hope you can fix it in a better way, thanks.

PR #55, opened 2017-05-08
If we receive an updated frame ensure we reset `cur` if required.

robholland:
Prior to this change `cur` can be stale, meaning the frame returned for
`WaitFrame(ctxt, cdptypes.EmptyFrameID)` will be out of date, and never updated
based on incoming events.


kenshaw:
Merged.

PR #64, opened 2017-06-01
Add support for event listeners

MrSaints:
Previously, the `Listen` method was not implemented, and it simply
returned `nil`. Consequently, any uses of the method will block
further code execution indefinitely.

This commit adds basic support for event listeners.
It creates an event listeners registry / pool on the target handler
which `Listen` registers against. Within the `processEvent` method,
it will loop through all the registered event listeners, and call
them if the event matches. The event listener channel will be given
the event / method type.

It can be used for creating a custom action that waits for the
page to load, for example:

```go
func WaitLoadEvent() cdp.Action {
    return cdp.ActionFunc(func(ctx context.Context, handler cdptypes.Handler) error {
        select {
        case <-handler.Listen(cdptypes.EventPageLoadEventFired):

        case <-ctx.Done():
            return ctx.Err()
        }
        return nil
    })
}
```

The above assumes the following import:

```
cdp "github.com/knq/chromedp"
```

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -41,6 +41,9 @@ type TargetHandler struct {
 	// qevents is the incoming event queue.
 	qevents chan *cdp.Message
 
+	// listeners is a pool of event listeners
+	listeners map[cdp.MethodType][]chan interface{}
+
 	// detached is closed when the detached event is received.
 	detached chan *inspector.EventDetached
 
@@ -247,6 +250,13 @@ func (h *TargetHandler) processEvent(ctxt context.Context, msg *cdp.Message) err
 		h.domWaitGroup.Wait()
 		go h.documentUpdated(ctxt)
 		return nil
+	}
+
+	// Trigger registered event listeners by giving the event
+	if listeners, ok := h.listeners[msg.Method]; ok {
+		for _, listener := range listeners {
+			listener <- ev
+		}
 	}
 
 	d := msg.Method.Domain()
@@ -668,7 +678,18 @@ func (h *TargetHandler) domEvent(ctxt context.Context, ev interface{}) {
 
 // Listen creates a listener for the specified event types.
 func (h *TargetHandler) Listen(eventTypes ...cdp.MethodType) <-chan interface{} {
-	return nil
+	out := make(chan interface{})
+
+	// Initialise listeners if it has not been initialised yet
+	if len(h.listeners) == 0 {
+		h.listeners = make(map[cdp.MethodType][]chan interface{})
+	}
+
+	// Register the listener
+	for _, eventType := range eventTypes {
+		h.listeners[eventType] = append(h.listeners[eventType], out)
+	}
+	return out
 }
 
 // Release releases a channel returned from Listen.


handler.go
```diff
@@ -249,6 +252,13 @@ func (h *TargetHandler) processEvent(ctxt context.Context, msg *cdp.Message) err
 		return nil
 	}
 
+	// Trigger registered event listeners by giving the event / method type
+	if listeners, ok := h.listeners[msg.Method]; ok {
+		for _, listener := range listeners {
+			listener <- msg.Method
```
rjeczalik: Shouldn't the listeners receive the `ev` event instead? I could imagine just method name without any payload may not be useful.

MrSaints: Yea, that's doable.

MrSaints: Done 👍

rjeczalik: It may be too much to ask, but what about Release() method? 😶

pwaller:
I'm a bit worried about races here.

Imagine this use case, where you want `DoSomething` to run after the page load fired:

```go
	err = c.Run(ctxt, cdp.Tasks{
		cdp.Navigate(`http://example.com`),
		WaitForPageLoadEventFired,
                DoSomething,

		cdp.Navigate(`http://example.com/other`),
		WaitForPageLoadEventFired,
                DoSomethingElse,
        })
```

How do you guarantee that the waiter in the hypothetical `WaitForPageLoadEventFired` is listening before the event fired? `knq/chromedp` is asynchronous, so there is nothing other than luck of timing preventing that from happening. If this happened, it would result in a hang, because the waiter missed the event and waits for something which will never come.

If you had some sort of timeout mechanism to work around this - then when `example.com/other` comes to load, if `example.com` is slow to load and times out, the second waiter might subsequently get the page load event intended for `example.com`, resulting in `DoSomethingElse` running *before* `example.com/other` has loaded. You would also risk such bugs if you did not consume events in the first page load, but only in subsequent loads - the subsequent loads could end up receiving events connected with the first page load.

These are two scenarios I think it is worth considering carefully before publishing this as a functional API. My experience of using Selenium hanging in the past has not been pleasant in the presence of such races. It would be great to escape them.

To avoid the former scenario, it seems necessary to set up the listener before it is needed, or to somehow replay past events into new listeners (but only those relevant to the listener, not ones from previous page loads, for example).

To avoid the second scenario, it might be necessary to have some mechanism which "resets" the stream of events when a navigation event happens, so that the subsequent waiter can't accidentally receive events intended for a previous page load.

Hope my suggestions here make some sense. I think this is a tough issue to solve.

pwaller:
I've implemented by suggestion in the above comment on top of the fix for #75 in my [`add-waiteventload`](https://github.com/pwaller/chromedp/commit/87ca8d1e42ce2d449d2eade137908409ee2cb773) branch. It enables this use case:

```go
	err = c.Run(ctxt, cdp.Tasks{
		cdp.Navigate(`http://example.com`),
		cdp.WaitEventLoad(),
                DoSomething,

		cdp.Navigate(`http://example.com/other`),
		cdp.WaitEventLoad(),
                DoSomethingElse,
        })
```

It seems to work and enable the use case I suggested in above. And I've attempted to fix the races and "event alignment problems" I discussed in the comment above - but of course there is always scope for more bugs. This stuff is quite hard.

The [patch](https://github.com/pwaller/chromedp/commit/87ca8d1e42ce2d449d2eade137908409ee2cb773) is not yet a generic way of waiting for events, it just solves the use-case where you want to wait for a page load before doing the next thing. It could be generalised to handle arbitrary events, but I figured it was best to figure out how to make it correct for just one important use case first.

pwaller:
I've [submitted a PR](https://github.com/knq/chromedp/pull/79) with my ideas in the above comments which potentially supersedes this PR. Comments welcomed!

PR #67, opened 2017-06-05
Update to latest protocol.json

vially:
This update should add support for intercepting network requests.


kenshaw:
I appreciate that you submitted this, but I can't accept PRs associated with the updated low level / auto generated protocols stuff as of today for "reasons". Apologies on that! Please note, however, that I just pushed out an almost identical change after updating to the latest protocol.json locally.

PR #69, opened 2017-06-06
Fix empty struct for CDP object types with no properties

### LINKED ISSUE #68: Unable to set extra HTTP headers
The `network` package's `SetExtraHTTPHeaders` method accepts a type of `Headers`. But it is an empty struct with no fields. It seems like it should really be a `map[string]interface{}` if there are no properties in the protocol definition. Otherwise, how can we set extra HTTP headers?

MrSaints:
`Headers` under the `network` package is an empty struct.

This is because the generator did not account for cases where
an object definition contains no properties.

In these cases, we can simply use `map[string]interface{}` so
that the type is treated as a generic key-value object.

This commit should allow extra HTTP headers to be set as
this was not possible with an empty type struct.

Closes #68

### DIFF
--- a/cdp/network/easyjson.go
+++ b/cdp/network/easyjson.go
@@ -51,7 +51,31 @@ func easyjsonC5a4559bDecodeGithubComKnqChromedpCdpNetwork(in *jlexer.Lexer, out 
 				if out.Headers == nil {
 					out.Headers = new(Headers)
 				}
-				(*out.Headers).UnmarshalEasyJSON(in)
+				if in.IsNull() {
+					in.Skip()
+				} else {
+					in.Delim('{')
+					if !in.IsDelim('}') {
+						*out.Headers = make(Headers)
+					} else {
+						*out.Headers = nil
+					}
+					for !in.IsDelim('}') {
+						key := string(in.String())
+						in.WantColon()
+						var v1 interface{}
+						if m, ok := v1.(easyjson.Unmarshaler); ok {
+							m.UnmarshalEasyJSON(in)
+						} else if m, ok := v1.(json.Unmarshaler); ok {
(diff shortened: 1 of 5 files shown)


kenshaw:
I'm reviewing this now; I appreciate you catching this, however I think that the way this should be fixed is by adding another "fixup" (`fixup/fixup.go`) to change the specific type declarations scanned from protocol.json. It's not that your approach is wrong, it's just that I believe we should be explicit with specific fixes.

MrSaints:
@kenshaw Cool, I'm happy to look into that if that's preferred.

kenshaw:
So -- after going through protocol.json, the only declared object types missing declared object properties are Network.Headers and Tracing.MemoryDumpConfig. Your approach might be the right approach regardless, however I am thinking a special type for Request.Headers might be warranted in order to provide some high-level wrappers around marshalling the std http.Header type. I need to do a couple more checks first on the low level protocol API to see what if there is any side-effect to sending anything other than map[string]string. If the expected JSON API has the Go equivalent type of map[string][]string (ie, compatible with http.Header), or if it is only map[string]string then the type you chose map[string]interface{} should be different (this is also why it should be an explicit fixup).

MrSaints:
@kenshaw Considering that it is used for setting request headers, I reckon it will mostly be string -> string | int | bool, and the value type can probably be conflated to string. I can't think of a case where we would want to set an array or a nested object as a value. That is, `map[string]string` should probably do it. The same cannot be said about `MemoryDumpConfig`.
I understand the preference for explicit fixes, but at the same time, it feels somewhat nasty to special case every type. Perhaps we can do both.

kenshaw:
I fixed this with a more general purpose fixup. Appreciate you finding / reporting this and submitting the original fix.

PR #76, opened 2017-06-21
Detect support for --headless and use if available

pwaller:
Chrome 59 gained support for `--headless`.

This PR makes the runner use this if `Headless` is supplied to the
runner, if it's available, falling back to the old method
(`headless_shell`) if it is unavailable.

Unfortunately it does not seem possible to get chrome to list which
command line options it supports. I could not find a way of detecting
the feature, so we instead just use a check on the version number, which
is not the preferred solution but does the job.

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -336,12 +336,21 @@ func Path(path string) CommandLineOption {
 	}
 }
 
-// Headless is the Chrome command line option to set the default settings for
-// running the headless_shell executable. If path is empty, then an attempt
-// will be made to find headless_shell on the path.
+// Headless uses the --headless command line option to run without a visible
+// window, if it is available (Chrome 59 onwards). If that does not work,
+// `headless_shell` is used as a fallback. If the path passed is an empty
+// string, an attempt is made to search for a suitable chrome/chromium binary to
+// invoke.
 func Headless(path string, port int) CommandLineOption {
+	var haveHeadless bool
 	if path == "" {
-		path, _ = exec.LookPath("headless_shell")
+		// Check for native --headless support, introduced in Chrome 59.
+		path = findChromePath()
+		haveHeadless = checkHeadlessSupport(path)
+		if !haveHeadless {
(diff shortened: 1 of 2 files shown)


pwaller:
For me, the above patch makes the tests pass without having anything other than chrome installed (no `headless_shell`).

pwaller:
I do not understand the test failures - they don't seem to be related to my change?

kenshaw:
I'm not sure what the benefit of this would be? I believe Chrome ignores any invalid command line parameters, so if you pass `--headless` and expect it to work and it doesn't, then you're on a version of Chrome / browser that doesn't support `--headless`.

kenshaw:
On another note, I'd like to keep "magic" to a minimum. Meaning, I don't think `chromedp` should be responsible for too much management of the browser process. There's a certain amount of invocation voodoo going on with the `runner` and `client` packages, for sure, but that's kind of the base minimum needed to "make it go." My concern with adding the additional logic in the command line option, is that there is likely a performance hit from `exec.LookPath` being called every time you launch a new `--headless` browser.

pwaller:
> I'm not sure what the benefit of this would be?

My main motivation is to make the tests work without having to install headless_shell.

> I believe Chrome ignores any invalid command line parameters, so if you pass --headless and expect it to work and it doesn't, then you're on a version of Chrome / browser that doesn't support --headless.

Except you then got a non-headless browser.

> On another note, I'd like to keep "magic" to a minimum.

Understandable.

> My concern with adding the additional logic in the command line option, is that there is likely a performance hit from exec.LookPath being called every time you launch a new --headless browser.

I understand where you're coming from, but I would expect this performance hit to be approximately nothing, given reasonable values for `$PATH`, and the "cost" of failed lookups would be quickly amortised by the OS's caches.

Also, my patch doesn't introduce any more lookpaths than were already present.

stephencheng:
@pwaller I guess this would be the treatment for my issue: https://github.com/knq/chromedp/issues/109

Good job

Issue #75, opened 2017-06-15
domEvent: timeout waiting for node 


rjeczalik:
This is more a question than a bug report.

When a dom event listener is about to wait for a node and the node gets concurrently removed (or its parent), WaitNode fails with timeout. The problem here is when that happens, the whole action handlers stale for 10 seconds. At least this is what I think happens. The behaviour is random, if WaitNode finishes before it's removed, then everything works as expected (I think also Go scheduler contributes to the racy behaviour).

Debug log for timeout scenario:

```
scrap.go:91: -> {"method":"DOM.childNodeInserted","params":{"parentNodeId":7,"previousNodeId":8,"node":{"nodeId":9,"backendNodeId":9,"nodeType":1,"nodeName":"DIV","localName":"div","nodeValue":"","childNodeCount":0,"attributes":[]}}}
scrap.go:91: -> {"method":"DOM.attributeModified","params":{"nodeId":9,"name":"id","value":"someotherid"}}
scrap.go:91: -> {"method":"DOM.childNodeInserted","params":{"parentNodeId":7,"previousNodeId":0,"node":{"nodeId":11,"backendNodeId":11,"nodeType":1,"nodeName":"BODY","localName":"body","nodeValue":"","childNodeCount":1,"attributes":["style","visibility: hidden; width: 0px; height: 0px; border: 0px; margin: 0px; background: none;"]}}}
scrap.go:91: -> {"method":"DOM.childNodeCountUpdated","params":{"nodeId":11,"childNodeCount":0}}
scrap.go:91: -> {"method":"DOM.childNodeRemoved","params":{"parentNodeId":7,"nodeId":9}}
scrap.go:91: -> {"method":"DOM.childNodeRemoved","params":{"parentNodeId":7,"nodeId":11}}
scrap.go:91: could not perform (attributeModified) operation on node 9 (wait node): timeout waiting for node `9`
scrap.go:91: could not perform (childNodeCountUpdated) operation on node 11 (wait node): timeout waiting for node `11`
```

I was thinking about fixing this (the 10s stales are really bugging me) by introducing some kind of context cancellation / listening channel, so we break out of waiting loop in WaitNode when the node we're waiting for gets removed.

Wdyt?

pwaller:
I'm seeing the same behaviour. At random, if a node is removed before it is waited on, we get a 10s hang.

I have tried @rjeczalik's patch, which seems to ameliorate the problem It does give an error which I think is not useful:

```
[...]

PR #77, opened 2017-06-21
Process events synchronously

pwaller:
This is a follow up to #75.

Previously, each event was processed in its own goroutine. This lead to
problematic order of event processing (e.g, node add|remove|modify where the
modify is trying to modify something that doesn't exist). It also lead to
goroutines waiting for things that would never subsequently happen.

Two significant changes in this PR:

* Do not run updates and event handlers each in their own goroutine, but
  instead run them in one goroutine in the order they came in. This
  avoids the race described in #75.

* Split the process queues into their own goroutines. This is necessary
  so that event handling can run commands. If all the queues are
  processed from the same goroutine, then when for exapmle when
  `documentUpdated()` runs,  it tries to run a `dom.GetDocument()`,
  which would deadlock because documentUpdated() is occupying the
  handler goroutine.

This patch does not represent a complete fix, because I believe it may make
a lot of locking and other concurrent machinary unnecessary. I don't have
time to explore that right now, but I don't think it's harmful as far as I
can tell.

I think there should in principle be a follow up PR to remove now-unnecessary
locks and waitgroups, but I'm not in a position to do those changes myself just
at the moment.

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -87,7 +87,15 @@ func (h *TargetHandler) Run(ctxt context.Context) error {
 	h.frames = make(map[cdp.FrameID]*cdp.Frame)
 	h.qcmd = make(chan *cdp.Message)
 	h.qres = make(chan *cdp.Message)
-	h.qevents = make(chan *cdp.Message)
+	// The events channel needs to be big enough to buffer as many events as
+	// we might recieve at once while processing one event. This is
+	// necessary because the event handling may do requests which require
+	// reading from the qmsg channel. This can't happen if the network
+	// handler is blocked writing to qevents. See discussion in #75. In
+	// practice I haven't seen the buffer get bigger than one or two. But we
+	// make it large just to be safe, and panic if it is ever full, rather
+	// than deadlocking.
+	h.qevents = make(chan *cdp.Message, 1024)
 	h.res = make(map[int64]chan *cdp.Message)
 	h.detached = make(chan *inspector.EventDetached)
 	h.pageWaitGroup = new(sync.WaitGroup)
@@ -156,7 +164,12 @@ func (h *TargetHandler) run(ctxt context.Context) {
 
 				switch {
 				case msg.Method != "":
-					h.qevents <- msg
+					select {
+					case h.qevents <- msg:
+					default:
+						// See discussion in #75.
+						panic("h.qevents is blocked!")
+					}
 
 				case msg.ID != 0:
 					h.qres <- msg
@@ -175,33 +188,58 @@ func (h *TargetHandler) run(ctxt context.Context) {
 		}
 	}()
 
-	var err error
+	var wg sync.WaitGroup
+	defer wg.Wait()
 
 	// process queues
-	for {
-		select {
-		case ev := <-h.qevents:
-			err = h.processEvent(ctxt, ev)
-			if err != nil {
-				h.errorf("could not process event %s: %v", ev.Method, err)
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for {
+			select {
+			case ev := <-h.qevents:
+				err := h.processEvent(ctxt, ev)
+				if err != nil {
+					h.errorf("could not process event %s: %v", ev.Method, err)
+				}
+			case <-ctxt.Done():
+				return
 			}
-
-		case res := <-h.qres:
-			err = h.processResult(res)
-			if err != nil {
-				h.errorf("could not process result for message %d: %v", res.ID, err)
-			}
-
-		case cmd := <-h.qcmd:
-			err = h.processCommand(cmd)
-			if err != nil {
-				h.errorf("could not process command message %d: %v", cmd.ID, err)
-			}
-
-		case <-ctxt.Done():
-			return
 		}
-	}
+	}()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for {
+			select {
+			case res := <-h.qres:
+				err := h.processResult(res)
+				if err != nil {
+					h.errorf("could not process result for message %d: %v", res.ID, err)
+				}
+			case <-ctxt.Done():
+				return
+			}
+		}
+	}()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for {
+			select {
+			case cmd := <-h.qcmd:
+				err := h.processCommand(cmd)
+				if err != nil {
+					h.errorf("could not process command message %d: %v", cmd.ID, err)
+				}
+
+			case <-ctxt.Done():
+				return
+			}
+		}
+	}()
 }
 
 // read reads a message from the client connection.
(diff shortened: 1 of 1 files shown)


pwaller:
/cc @rjeczalik 

There is also this [patch](https://github.com/rjeczalik/chromedp/commit/1d3ee5b2625b76a2e3f77b8ec1be7512cfb8d2cb) to what I've done to ignore waits on empty nodes. I haven't hit that case myself and the commit has a comment stating that it might not be the best way to handle it. It would be useful to have a reproducer for it so that we can make a better fix for it.

I neglected to mention in the PR body 00b69aa which ignores EventFrame state changes. This is necessary because those events seem to genuinely refer to FrameIDs which don't yet (and might never) have a corresponding `cdf.Frame` object associated with them. So for now I simply drop those, since it seems not possible to use those state changes correctly in the current implementation.

rjeczalik:
>  It would be useful to have a reproducer for it so that we can make a better fix for it.

I think the bug manifests itself when we cdp.Navigate to a different url while current frame is still being loaded - it may somehow invalidate old nodes thus 0 received via cdp events. I did not debug it, this is just my wild guess.

pwaller:
Are you able to come up with some minimal setup which occasionally triggers ityou can share?

rjeczalik:
@pwaller If I got couple of spare minutes I'm going to reproduce this and create separate issue/PR. For now just sharing observations if it happens someone will get around solving this problem faster.

pwaller:
I think I've found the root cause of the waiting on empty nodes. It's actually because `documentUpdated` puts the nodes on the wrong frame. I've struggled to find a reliable simple reproducer - loading a Google document does the trick.

I think the underlying problem here is with the way that `h.cur` works. The problem is that we want to update it at the point of the navigation, but at that point we don't yet have a `*cdp.Frame`. This only comes later.

It might be better if the `h.cur` mechanism didn't store the frame object but just the ID - and the actual user of the frame then calls `WaitFrame`. I'll give this a try if I get a moment.

```
2017/06/30 20:22:28 cmd 9/"Page.navigate"
2017/06/30 20:22:29 Got result for 9/""
2017/06/30 20:22:29 Event *page.EventFrameStartedLoading
2017/06/30 20:22:29 SetActive waiting for frame "19500.1" // caused by navigation
2017/06/30 20:22:29 Event *page.EventFrameNavigated
2017/06/30 20:22:29 Event *dom.EventDocumentUpdated
2017/06/30 20:22:29 WaitFrame(id="") = 0xc420086140
2017/06/30 20:22:29 documentUpdated operating on 0xc420086140 starts
2017/06/30 20:22:29 cmd 10/"DOM.getDocument"
2017/06/30 20:22:29 WaitFrame(id="19500.1") = 0xc4224d6320 // SetActive's wait on the frame finally completes.
2017/06/30 20:22:29 SetActive sets h.cur = 0xc4224d6320 // caused by navigation
2017/06/30 20:22:29 Got result for 10/""
2017/06/30 20:22:29 documentUpdated operating on 0xc420086140 completes // Old h.cur
2017/06/30 20:22:29 Event *dom.EventChildNodeCountUpdated
2017/06/30 20:22:29 WaitFrame(id="") = 0xc4224d6320 // New h.cur
2017/06/30 20:22:30 WaitNode spinning on 0xc4224d6320
```

kenshaw:
So, we can't really make this synchronous. However, there absolutely is a problem with the implementation, and I think you're partially correct that it should store the frame ID and then essentially wait until that frame i present. What I'm going to do is split out the code from the current handler, into a separate manager type. I don't have a fully formed idea yet on this, but it'll have significant implications for the API in general. I'll play around with it, and compare some different approaches / styles to see what the best solution would be.

pwaller:
> So, we can't really make this synchronous.

Could you elaborate on the reasoning? From my research I disagree (to a degree). This statement by itself makes it hard to give any useful feedback.  I also fear we may talk cross purposes unless I can better understand your motivations.

There are some things that can be done asynchronously, but I think specifically that processing every incoming event in an arbitrary order is  complicating things and hiding several bugs such as the one in my comment above. These bugs make it quite hard to get started with ChromeDP - see for example #80, #75, [this comment](https://github.com/knq/chromedp/issues/82#issuecomment-312282046). I also believe that there are often goroutines spinning waiting for elements which will never appear because of these issues.

I believe that certain heavy lifting can be done in parallel (if that is the motivation - and if it is 
 worth it) while still doing the event processing in-order, with some care. I think the net result could end up being easier to reason about than the current approach, and possibly faster.

derlaft:
@knq, whats your status about it? Without this patch I get tons of that errors and lots of seconds of lags.

t-blake:
I have been using this patch for several days and it has eliminated the lags and missing node errors that I was experiencing with the master branch. 
Many thanks @pwaller

stephencheng:
@pwaller A thumb up of your merge request to improve the efficiency of 5000 times! Great Job!

jcrussell:
I made some cleanups and merged to master: https://github.com/jcrussell/chromedp/commit/cf34da2947650178ca6d1550c3c02de9b6750b60.

Seems to be working great!

bcspragu:
I'm currently evaluating knq/chromedp and tebeka/selenium for some browser automation tasks. Without this pull request, a basic 'Hello World' style example takes ~30s in chromedp and ~2s in selenium, despite chromedp being the 'lightweight' approach. 

The example in ChromeDP:
```go
err = c.Run(ctx, cdp.Tasks{
  cdp.Navigate("http://webscraper.io/test-sites/e-commerce/allinone"),
  cdp.Click("//a[text()[contains(., 'Computers')]]", cdp.NodeVisible),
  cdp.Click("//a[text()[contains(., 'Laptops')]]", cdp.NodeVisible),
  cdp.Nodes(".thumbnail .title", &titles, cdp.ByQueryAll),
})
```

And now in Selenium:

```go
if err := wd.Get("http://webscraper.io/test-sites/e-commerce/allinone"); err != nil {
  return nil, err
}

el, err := wd.FindElement(selenium.ByXPATH, "//a[contains(text(), 'Computers')]")
if err != nil {
  return nil, err
}

if err := el.Click(); err != nil {
  return nil, err
}

el, err = wd.FindElement(selenium.ByXPATH, "//a[contains(text(), 'Laptops')]")
if err != nil {
  return nil, err
}

if err := el.Click(); err != nil {
  return nil, err
}

els, err := wd.FindElements(selenium.ByCSSSelector, ".thumbnail .title")
if err != nil {
  return nil, err
}
```

Without this patch:

```
$ go test --bench=. --benchtime=30s

goos: linux
goarch: amd64
pkg: github.com/bcspragu/auto-test
BenchmarkChromeDP-12                   1        32088692546 ns/op
BenchmarkSelenium-12                  20        2212985810 ns/op
PASS
ok      github.com/bcspragu/auto-test   80.671s
```

With this pull request applied, the results are more even:

```
$ go test --bench=. --benchtime=30s

goos: linux
goarch: amd64
pkg: github.com/bcspragu/auto-test
BenchmarkChromeDP-12                  20        1993346253 ns/op
BenchmarkSelenium-12                  20        2238880901 ns/op
PASS
ok      github.com/bcspragu/auto-test   91.212s
```

I'd like to use (and contribute to) the ChromeDP library, because the API is far easier to work with and doesn't require Java/Selenium, but being 15x slower is a non-starter. Is there some way to reconcile a fix here?

stephencheng:
@bcspragu it's just a matter of time this fix will be merged back. I am in the same boat and currently use the temp fix, till then, it will only take millseconds for the job

stephencheng:
@kenshaw seems this pull request was not merged back to master branch.

Understand that you have the project split. What's your plan or thought to adopt this pull request? I wish it could be merged sooner than late as it helps the performance

PR #78, opened 2017-06-21
Support element screenshots for ScaleFactor != 1

pwaller:
This enables making screenshots of zoomed elements, for example to make
high-dpi screenshots.

### DIFF
--- a/query.go
+++ b/query.go
@@ -428,13 +428,17 @@ func Screenshot(sel interface{}, picbuf *[]byte, opts ...QueryOption) Action {
 			return ErrInvalidBoxModel
 		}
 
-		// scroll to node position
+		var pixRatio float64
+		err = EvaluateAsDevTools(getWindowDevicePixelRatioJS, &pixRatio).Do(ctxt, h)
+		if err != nil {
+			return err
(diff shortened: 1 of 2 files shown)


pwaller:
To make use of this I have an ActionFunc like this:

```go
func setViewportAndScale(w, h int64, scale float64) cdp.ActionFunc {
	return func(ctxt context.Context, ha cdptypes.Handler) error {
		err := emulation.SetVisibleSize(w, h).Do(ctxt, ha)
		if err != nil {
			return err
		}
		sw, sh := int64(float64(w)/scale), int64(float64(h)/scale)
		err = emulation.SetDeviceMetricsOverride(sw, sh, scale, false, false).WithScale(scale).Do(ctxt, ha)
		if err != nil {
			return err
		}
		return nil
	}
}
```

This allows you to make a screenshot which is `(w,h)` in size in the resulting PNG, with the given zoom factor applied.

kenshaw:
What you have should be part of the "navigation" actions. I'll look at adding something like this. Apologies for not thinking about DPI/scaling factor with the original screenshot actions.

PR #79, opened 2017-06-21
Add WaitEventLoad action

pwaller:
This is a follow up to [my comments on #64](https://github.com/knq/chromedp/pull/64#issuecomment-309056121). It probably doesn't make sense to merge this PR before #77 is completed, but I wanted to put this out there for feedback and review.

I don't expect this PR to be merged as is, it's just a concept at this stage. I think the obvious thing to do is to generalise it to enable waiting on any kind of event. I don't yet know how to do this in particular because I'm not sure what the potential use cases are.

If this PR or a variant of it is merged, I would recommend removing the existing [`Listen`/`Release` stubs](https://github.com/knq/chromedp/blob/4cc4ba1ca23fb669e5ec66cda5c2bcdc3c10f672/handler.go#L668-L676).

The main approach with this PR is to put careful consideration into race conditions, and trying to marry up the load event with the appropriate navigation event. I'm not certain that this implementation is correct or the best form to put this in. I just wanted to get the conversation started. There are likely other use cases which are not convered, e.g, navigation events which are caused by things other than a `Navigate.Do` such as form submission, etc, which might be a good idea to consider in more detail.

Example use case:

```go
cdp.Tasks{
	cdp.Navigate("http://localhost:8080/foo"),
	cdp.WaitEventLoad(),
	cdp.CaptureScreenshot(&buf),
	saveScreenshot("foo.png"),

	cdp.Navigate("http://localhost:8080/bar"),
	cdp.WaitEventLoad(),
	cdp.CaptureScreenshot(&buf),
	saveScreenshot("bar.png"),

	...
}
```

I have been using this for a few days now, and I appear to reliably and quickly get screenshot of completed pages. This approach has the advantage over the traditional `WaitVisible` that it doesn't require polling (instead it is reacting to the browser declaring that the page has loaded) - so it's a bit faster, and it doesn't require page-specific knowledge about what can be waited on.

Do not merge before #77.
---

### DIFF
--- a/actions.go
+++ b/actions.go
@@ -59,3 +59,15 @@ func Sleep(d time.Duration) Action {
 		return nil
 	})
 }
+
+func WaitEventLoad() Action {
+	return ActionFunc(func(ctxt context.Context, h cdp.Handler) error {
+		select {
+		case <-h.WaitEventLoad():
+		case <-ctxt.Done():
+			return ctxt.Err()
+		}
+		return nil
+	})
+
+}
--- a/cdp/cdp.go
+++ b/cdp/cdp.go
@@ -1406,6 +1406,9 @@ type Handler interface {
 	// WaitNode waits for a node to be available.
 	WaitNode(context.Context, *Frame, NodeID) (*Node, error)
 
+	ResetEventLoad()
+	WaitEventLoad() <-chan struct{}
+
 	// Execute executes the specified command using the supplied context and
 	// parameters.
 	Execute(context.Context, MethodType, easyjson.Marshaler, easyjson.Unmarshaler) error
--- a/handler.go
+++ b/handler.go
@@ -44,6 +44,9 @@ type TargetHandler struct {
 	// detached is closed when the detached event is received.
 	detached chan *inspector.EventDetached
 
+	// loaded is fired when the page load event is received.
+	loaded chan struct{}
+
 	pageWaitGroup, domWaitGroup *sync.WaitGroup
 
 	// last is the last sent message identifier.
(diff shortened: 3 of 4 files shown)


rjeczalik:
Excuse my unrelated comment, but going through travis output for this PR I've noticed this failure:

```
chromedp_test.go:44: target handler error: could not retrieve document root for 8948.1: context canceled
```

In my project I've noticed this (`context canceled`) a number of times, looks like frame loading gets interrupted when it's too slow (the hardcoded timeouts are backfiring?). Again, I did not debug this, just an observation.

pwaller:
@rjeczalik I suspect this might be fixed in #77 - it's likely a goroutine for an event asynchronously waiting for something, but the overall work completes so the context is cancelled. I would be interested if you can reproduce it with the patches in #77 applied, I suspect not.

pwaller:
I think the client code could look something like this, and potentially solve all the different race conditions, so far as I can reason:

```go
cdp.Tasks{
        cdp.WithListener(func(pageLoad <-chan *page.EventLoadEventFired) Action {
                return cdp.Tasks{
                        cdp.Navigate(url),
                        cdp.Wait(e),
                }
        }),
        // Do screenshot, etc.
}
```

`WithListener(makeAction {}interface) Action` brackets the code which will wait on the event. It serves the purpose of enabling chromedp to enforce the important invariants:

* The event listener is in place before the actions which might cause the event (e.g, the `cdp.Navigate` causing `*page.EventLoadEventFired`).
* At the end of the closure the channel can be closed so that any asynchronous code can't receive events destined for future `WithListener`.

The `makeAction` function can be a function receiving any `*EventType`. Reflection can determine what kinds of events are being listened on. Variadic closures could enable waiting on many event types at once.

`Wait` is a convenience action which simply unblocks when the channel is read. But the user is free to write client code which waits on the channel and actually sees the underlying type-safe `e *EventType`. `Wait`
 or perhaps `MustWait `could return an error if the channel is closed without receiving anything, to prevent bugs where the event is not received before the code moves on to other things which could generate the event. There could be a timeout'd variant `WaitTimeout(e, duration)`.

The above solution seems neatly composable, for example the above could be written in client code as `func NavigateWait(url string) Action` which does the navigation and waits for the load event to complete. By forcing the user to pass their code into `WithListener` if they want to receive events, I think many types of bugs can be prevented.

It would be even better if we could find a way to raise an error in this erroneous example below, but I haven't yet thought of a clean way of doing that. One style of solution would be to assert that something else must happen before the wait, but of course that isn't possible if you pass in events...

```
cdp.Tasks{
        cdp.Navigate(url),
        cdp.WithListener(func(pageLoad <-chan *page.EventLoadEventFired) Action {
                return cdp.Wait(e)
        })
        // Do screenshot, etc.
}
```

kenshaw:
Please wait for the newest revision of chromedp -- I've taken a different approach that should solve this, as I had a different need for this during the network code refactoring I've had to do. I greatly appreciate the contributions that have been done though!

yamoe:
is done?

smcduck:
WithListener is not available now

PR #80, opened 2017-06-22
Fixes infinite loop (and consequentially spiked CPU usage) on Ctx Timeout

JalfResi:
Fixes infinite loop (and consequentially spiked CPU usage) on ctx timeout. Noticed this issue when trying to limit the max execution time of a crawl job by using Context.WithTimeout().


vially:
I'm being affected by this too. Thanks for the patch.

Any particular reason this was not merged yet? //cc @kenshaw

kenshaw:
Apologies on the delay. Merged.

JalfResi:
Awesome thanks!

PR #85, opened 2017-07-01
Attempt to fix unit tests.

ranch:


kenshaw:
@ranch let's discuss more about how I'm planning on fixing this once and for all. I've identified a logic/design problem that will hypothetically fix this. I believe the problems with travis are simply the same WaitNode locking issues being seen elsewhere.

PR #104, opened 2017-08-13
Add extra configurations

igsky:
These changes provide couple of useful features:

1. Enable necessary chrome domains only or enable some extras alongside default ones. You can check out usage in "scrape" example.
2. MsgHooks that allow you to intercept messages from chrome and do something with them. Quite a basic use case is to capture http status code, which can be quite important for scraping. Check out "scrape" example.
3. Added Config struct. MsgHooks and extra chrome domains add new variables to pass to NewTargetHandler, which is a bit messy, so I packed them up for ease of use and more concise code.

None of these changes should break any pre-existing code.


kenshaw:
These are too many different, unrelated changes in one pull request. Please separate these and flatten/rebase each into separate PRs. I'm definitely willing to merge many of these, but these need to be properly separated, "atomic" and clean.

igsky:
It's literally my first PR, so it's all over the place, sorry about that. I'll fix it ASAP.

PR #112, opened 2017-08-26
Splitting options into separate file

igsky:
There are quite a bunch of options at the moment, and more to come. So I think it's reasonable to split them into separate file for ease of use.


igsky:
Looks like my previous PR's failed due to some kind of external reason (server down or something like that).

I'm trying to keep commits atomic and easy to understand. Looking forward to hear feedback.

PR #146, opened 2017-11-29
Fix events being processed out of order.

jcrussell:
TargetHandler was firing off Goroutines for each page and dom event
which would run in arbitrary order. Force events to process in order by
putting them in a channel and having a single Goroutine read each.

@kenshaw @bbeck

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -22,6 +22,12 @@ import (
 	"github.com/knq/chromedp/client"
 )
 
+// event wraps a page or dom event with a context.
+type event struct {
+	ctxt context.Context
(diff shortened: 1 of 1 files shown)


jcrussell:
Seeing these errors now:

```
could not perform (childNodeCountUpdated) operation on node X (wait node): timeout waiting for node `X`
```

It holds up the program for a bit and then, after enough timeouts, it keeps going. Going to try making the `pageEvents` buffered.

jcrussell:
I hacked on this some more before finding that I had almost reimplemented #77.

Issue #148, opened 2017-12-01
grab-headless_shell.sh no longer functional


lrstanley:
just a heads up, looks like the account behind the binary-distributed compiled headless shell versions is no longer a thing.

```
liam:~$ curl -s https://storage.googleapis.com/docker-chrome-headless/latest.txt
<?xml version='1.0' encoding='UTF-8'?><Error><Code>AccountProblem</Code><Message>Your Google account is not currently enabled for this operation. Please check https://console.developers.google.com/billing to see if you have a past due balance or if the credit card (or other payment mechanism) on your account is expired. You can find additional information at https://developers.google.com/storage/docs/signup</Message><Details>The billing account for the requested project is disabled in state 'delinquent'</Details></Error>
```

not sure if there are any alternatives for easily grabbing a compiled version of the headless shell. looks like the `knqz/chrome-headless` docker image may be the best suggestion for folks at this time.

kenshaw:
Too many people were using that for retrieving headless_shell. The bill last month was over $1,000 USD. I would have left it up if it were less than $100, but as it is, there are a bunch of bots out there that were setup to use the published artifacts. I haven't had time to switch to a different location yet.

lrstanley:
Yikes!

Issue #101, opened 2017-08-05
Shutdown doesn't quit headless_shell


kmanley:
Repro case below. Shutdown does not seem to shut down headless_shell, so the Wait hangs.

```go
package main

import (
	"context"
	"log"
	cdp "github.com/knq/chromedp"
	"github.com/knq/chromedp/runner"
)

func main() {
	var err error

	ctxt, cancel := context.WithCancel(context.Background())
	defer cancel()

	log.Println("starting headless_shell")
	c, err := cdp.New(ctxt, cdp.WithLog(log.Printf),
		cdp.WithRunnerOptions(runner.Headless("/home/kevin/Downloads/headless_shell/headless_shell", 8222)))
	if err != nil {
		log.Fatal(err)
	}

	log.Println("shutting down...")
	err = c.Shutdown(ctxt)
	if err != nil {
		log.Fatal(err)
	}

	log.Println("waiting for headless_shell to exit...")
	err = c.Wait()
	if err != nil {
		log.Fatal(err)
	}

	log.Println("done")
}
```

pwaller:
I think this is a duplicate of #81.

kmanley:
@pwaller thx

mvdan:
Duplicate of https://github.com/chromedp/chromedp/issues/81.

PR #149, opened 2017-12-04
Update headless runner to use --headless

bcspragu:
The headless_shell binary is no longer the recommended way to run Chrome headless. Chromium documentation [1] points to using Chrome with the --headless flag, and per #148, the headless_shell binary isn't available for download anymore.

To make the tests run, I was aliasing headless_shell to `chrome --headless`, which was causing an issue similar to #101. With this change, the tests properly clean up resources on Shutdown and pass, with no headless_shell in my path.

[1] https://chromium.googlesource.com/chromium/src/+/lkgr/headless/README.md


kenshaw:
I appreciate the work you did -- I was not aware that Travis had a chrome addon. I was not able to merge your PR for a couple reasons: 1) not rebased off master to a single commit, 2) didn't actually fix issues with running unit tests. I've fixed the master branch so that unit tests run a bit better, and so that there are more options for controlling how they are ran.

Thanks!

PR #161, opened 2017-12-25
handler: use child context

### LINKED ISSUE #160: goroutine  leak in TargetHandler
When the context is done,  there will be no receiver on the h.qcmd channel in https://github.com/knq/chromedp/blob/9424dc57d6f066b60b2d1d1af2e265b09f027315/handler.go#L175

```go
	for {
		select {
		case ev := <-h.qevents:
			err := h.processEvent(ctxt, ev)
			if err != nil {
				h.errorf("could not process event %s: %v", ev.Method, err)
			}

		case res := <-h.qres:
			err := h.processResult(res)
			if err != nil {
				h.errorf("could not process result for message %d: %v", res.ID, err)
			}

		case cmd := <-h.qcmd:
			err := h.processCommand(cmd)
			if err != nil {
				h.errorf("could not process command message %d: %v", cmd.ID, err)
			}

		case <-ctxt.Done():
			return
		}
	}
```

then send message to h.qcmd channel on the other side will stuck for ever:
https://github.com/knq/chromedp/blob/9424dc57d6f066b60b2d1d1af2e265b09f027315/handler.go#L347

```go
	case h.qcmd <- &cdp.Message{
		ID:     id,
		Method: commandType,
		Params: paramsBuf,
	}:
```

Darren:
this fixes #160


kenshaw:
Similarly to my comment on the other PR, I'll address this shortly. Thank you for contributing!

Darren:
chromedp is still leaking memory. Is there any update on this?

 @kenshaw

Darren:
Any update on this? we switch to your master branch, and our application eats up 10GB of memory as I commented on #160

PR #162, opened 2017-12-26
sel: use buffered chan for error

Darren:
using unbuffered chan both 

https://github.com/chromedp/chromedp/blob/9424dc57d6f066b60b2d1d1af2e265b09f027315/sel.go#L88

and 

https://github.com/chromedp/chromedp/blob/9424dc57d6f066b60b2d1d1af2e265b09f027315/sel.go#L107

will stuck when called from Do()

```go
	var err error
	select {
	case err = <-s.run(ctxt, h):
	case <-ctxt.Done(): // when this is selected first , run can not send error  because of  no receiver
		err = ctxt.Err()
	}

	return err 
```


kenshaw:
I appreciate bringing this to my attention. I just finished reworking the repository structure / layout, and will address this shortly.

Darren:
@kenshaw 
I've have added a test case for this based on your new repo.
We found the leak while benchmarking our application which is based on chromedp, the test case I added just create context then immediately cancel it before passing to chromedp to Run.

PR #231, opened 2018-09-07
Add optional network events listener

dmitry-vovk:
Add support for getting network events:
```go
chrome, _ := chromedp.New(ctx, options...)
chrome.Run(ctx, chromedp.ActionFunc(func(_ context.Context, h cdp.Executor) error {
	go func() {
		for evt := range h.(*chromedp.TargetHandler).NetEvents() {
			switch e := evt.(type) {
			case *network.EventRequestWillBeSent:
				log.Printf("Sending %s request to %s", e.Request.Method, e.Request.URL)
			case *network.EventResponseReceived:
				log.Printf("Got %d response from %s", e.Response.Status, e.Response.URL)
			}
		}
	}()
	return nil
}))
```

### DIFF
--- a/chromedp.go
+++ b/chromedp.go
@@ -14,10 +14,9 @@ import (
 	"sync"
 	"time"
 
+	"github.com/BrightLocal/chromedp/client"
+	"github.com/BrightLocal/chromedp/runner"
 	"github.com/chromedp/cdproto/cdp"
-
-	"github.com/chromedp/chromedp/client"
-	"github.com/chromedp/chromedp/runner"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


UtopleMan:
Hey @dmitry-vovk do you have an example on how to retreive the response headers? I can see the network event listener has been added but a bit unsure where to pull the events from..

dmitry-vovk:
> Hey @dmitry-vovk do you have an example on how to retreive the response headers? I can see the network event listener has been added but a bit unsure where to pull the events from..

You will have both request and response headers in your `Response` field of `network.EventResponseReceived` variable. Named `Headers` and `ResponseHeaders` respectively.

PR #240, opened 2018-10-26
Adding query timeout feature proposition

mawasak:
Timeout may be set individually in every call using QueryOption
e.g.
chromedp.Query(sel, chromedp.QueryTimeout(1*time.Second)).Do(ctx, ex)

### DIFF
--- a/chromedp.go
+++ b/chromedp.go
@@ -33,6 +33,10 @@ const (
 
 	// DefaultPoolEndPort is the default end port number.
 	DefaultPoolEndPort = 10000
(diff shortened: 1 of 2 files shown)


mvdan:
Thanks for the patch. We're designing v2 now, and we believe there are better ways to handle this. In particular, practically all internal timeouts should be replaced with `context.Context`. So the user should be able to attach a context (and thus a timeout or a cancel func) to any action, not just a Query.

Stay tuned!

PR #246, opened 2018-11-07
Ability to configure defaults timeouts and checks duration

fydrah:
As a workaround for too many "timeout waiting for node XX", give the ability to set custom timeouts.

This **does not solve any issue**, but it extracts some hard coded timeout values.


Northern-Lights:
Referencing #248

mvdan:
Thanks for your patch! Please see my reply to https://github.com/chromedp/chromedp/issues/248#issuecomment-465543973. We'll overhaul the API for v2, and we think properly using `context.Context` will give users all the timeout configurability needed without bloating and complicating the API.

PR #251, opened 2018-11-16
Expose browser events via a hook function

dennwc:
Currently, there is no way to stream console output from the browser to Go.

This PR adds an `OnEvent` method to `TargetHandler` that sets a hook for incoming browser events. This allows the user to inspect events and react not only to console output but on other kinds of events as well.

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -42,6 +42,7 @@ type TargetHandler struct {
 
 	// qevents is the incoming event queue.
 	qevents chan *cdproto.Message
+	onEvent EventFunc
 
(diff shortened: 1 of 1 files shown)


dennwc:
CI failure seems unrelated. You might want to allow failures on `tip` in Travis.

mvdan:
Thanks for the patch. We're in the middle of a redesign for v2, so the API and internal structure might change a bit. This seems like a good idea right now, but we'd like to hold off on adding new features until v2 is done.

PR #254, opened 2018-11-22
pool: error quickly if we find a port in use

### LINKED ISSUE #253: Confusing error if allocating a process in a pool hits a port in use
This had a couple of engineers scratching their heads for a good half an hour, since the initial error didn't point to the source of the problem.

Here's a small reproducer, using chromedp 0.1.2:

```
package main

import (
	"context"
	"log"
	"net"

)

func main() {
	// take the first port in the default chromedp pool port range
	l, err := net.Listen("tcp", ":9000")
	if err != nil {
		log.Fatal(err)
	}
	defer l.Close()

	ctxt, cancel := context.WithCancel(context.Background())
	defer cancel()

	pool, err := chromedp.NewPool()
	if err != nil {
		log.Fatal(err)
	}

	c, err := pool.Allocate(ctxt)
	if err != nil {
		log.Fatal(err)
	}
	if err := c.Release(); err != nil {
		log.Fatal(err)
	}
}
```

```
$ go run main.go
2018/11/22 12:31:19 ERROR: pool could not connect to 9000: timeout waiting for initial target
2018/11/22 12:31:19 timeout waiting for initial target
exit status 1
```

It's obvious that something went wrong with connecting to the remote debugging port, but it's not obvious why. It seems like starting up Chrome went fine.

Of course, the underlying issue is that the port is in use, and Chrome will happily keep running even when it can't bind to that address. I'll submit a PR shortly to error sooner with a clearer message, also adding a test.

mvdan:
Before the fix, the added test would give a Pool.Allocate error like:

	pool could not connect to 9000: timeout waiting for initial target

The actual underlying error, which can only be seen if one inspects
chrome's stderr, is that it failed to bind to the debugging protocol
port if it was already in use.

This is of course an issue with the environment that chromedp is being
run under, since it was given a port range that wasn't available.
However, the confusing error can lead to developers wasting their time
instead of spotting the error quickly.

Unfortunately, there doesn't seem to be a way to have Chrome exit
immediately if it can't bind to the given port. So, instead of relying
on it, check if the current process can bind to the port first.

Add a test too, where we grab the first port in the pool range, and
check that we get an error that's not confusing.

Fixes #253.

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"net"
 	"sync"
 
 	"github.com/chromedp/chromedp/runner"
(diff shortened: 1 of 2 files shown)


mvdan:
The CI failure seems to be goveralls - unrelated.

PR #259, opened 2018-12-03
Event handler: CDP option for exposing events to a consumer

Northern-Lights:
This change allows a consumer to create a `chromedp.CDP` with an optional channel (using a function defined as `func WithEvents(ev chan<- *CDPEvent) Option`) for consuming events.

A `CDPEvent` struct was created and used as the container for the event (`cdproto.Message`) so that the target ID can accompany it, telling us what target (tab) the event came from.

Subscribing to the event stream is completely optional and will not affect current users.

### DIFF
--- a/chromedp.go
+++ b/chromedp.go
@@ -54,7 +54,7 @@ type CDP struct {
 	cur cdp.Executor
 
 	// events is the channel for piping cdproto events from handler to consumer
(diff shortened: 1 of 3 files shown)


mvdan:
#251 was filed before this one and seems like a simpler approach, so I'm going to close this one in favor of that one.

PR #271, opened 2019-01-24
use buffered chan for h.detached

zjj:
this patch fixed goroutine leaks when processing detached event.

224 // processEvent processes an incoming event.
225 func (h *TargetHandler) processEvent(ctxt context.Context, msg *cdproto.Message) error {
...
235 
236     switch e := ev.(type) {
237     case *inspector.EventDetached:
238         h.Lock()
239         defer h.Unlock()
240         h.detached <- e   **the socket may have closed, event pushed goroutine returns, blocking here.**


zjj:
this patch fixed goroutine leaks when processing detached event.

Issue #274, opened 2019-02-03
Send SIGTERM on os/archs other than osx (eg raspberrypi)


eginez:
First off awesome work.

Just wanted to let you know that I haven been testing this on an rasberrypi and it works almost out of the box(so far at least), except for a few tweaks I had to do 
 - startup time needs to be longer
 - the sigterm signal needs to be sent as well

I might send a PR with such changes if you are interested

leidegre:
Great.

I've been considering this as well. Would be great if we could get it to work with a switch or similar if it causes issues on other environment.

kenshaw:
Yes, a PR would be appreciated.

killianbrackey:
I put a quick PR together that has been tested for linux.  I didn't add a default switch case with a personal lack of familiarity on the need for other os/archs.

PR #276, opened 2019-02-25
Added SIGTERM to linux systems on shutdown

killianbrackey:
Running this on a Ubuntu 18, I noticed that the .Shutdown() function wasn't completing and came across #274 .  This switch has been tested on macOS and Ubuntu 18.

Closes https://github.com/chromedp/chromedp/issues/274

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -241,12 +241,15 @@ func (r *Runner) Shutdown(ctxt context.Context, opts ...client.Option) error {
 		}
 	}
 
-	// osx applications do not automatically exit when all windows (ie, tabs)
+	// osx and linux applications do not automatically exit when all windows (ie, tabs)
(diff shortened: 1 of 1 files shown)


runner/runner.go
```diff
@@ -245,8 +245,11 @@ func (r *Runner) Shutdown(ctxt context.Context, opts ...client.Option) error {
 	// closed, so send SIGTERM.
```
mvdan: This comment needs to be updated, as the code no longer only handles OSX.

killianbrackey: I resolved in my latest commit, but let me know if there is anything else needed for this!

killianbrackey:
@mvdan I force pushed the changes to keep the commit history clean.  It looks like this didn't fix up the "Requires changes" to push this back to you so you may need to do that on your end.

PR #283, opened 2019-03-20
fixed always waiting navigate action when context canceled

Code-Hex:
I have been fixed issue that always waiting navigate action after context canceled.

This code to reproduce this issue.

thanks :D

```go
// create context
ctx, cancel := context.WithCancel(context.Background())

// create chrome instance
chrome, err := chromedp.New(ctx,
	chromedp.WithRunnerOptions(runner.Headless),
)
if err != nil {
	return err
}

cancel()

time.Sleep(time.Second)

fmt.Println("finished sleep")

// infinity wait
return chrome.Run(ctx, chromedp.Tasks{
	chromedp.Navigate("https://google.com"),
})
```

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -346,10 +346,14 @@ func (h *TargetHandler) Execute(ctxt context.Context, methodType string, params 
 	h.resrw.Unlock()
 
 	// queue message
-	h.qcmd <- &cdproto.Message{
+	select {
+	case h.qcmd <- &cdproto.Message{
 		ID:     id,
 		Method: cdproto.MethodType(methodType),
 		Params: paramsBuf,
+	}:
+	case <-ctxt.Done():
+		return ctxt.Err()
 	}
 
 	errch := make(chan error, 1)


mvdan:
Hi, and thanks for your patch!

We're currently working on a full refactor of the API and internal design. I'm fairly certain that you wouldn't have these issues with the new version; it's context-based at core, so it should be possible to stop it or cancel it at any point.

Consider having a look at https://github.com/chromedp/chromedp/tree/refactor, and trying to reproduce the same issue once we release an alpha in the coming weeks. I don't think it's worth trying to fix issues in the old codebase; we should instead focus on the better design from the ground up :)

Code-Hex:
@mvdan That's great! 

However we are using current version as production stage. we hope this patch will be merged into the current version.

This patch is **only single line**, and I think it will not affect refactoring.

So please merge and release it :bow:

mvdan:
This piece of code is entirely gone from the upcoming version, and it would take some time to properly test the change for regressions, so I'm sorry but I disagree about its effect on the refactor.

I appreciate that you're using the stable version on production and want this patch, but you can do that without requiring that we merge and release the patch. For example, both Go modules and older dependency managers like dep allow replacing one dependecy with a forked git repository.

It's unfortunate that the master branch is in this transitional state, but we plan to end that soon. Hope that makes sense.

Code-Hex:
@mvdan Do you mean that we should use forked version using dep or modules?

mvdan:
Yes. See https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive, for example. It should be easy to set up, and you can apply as many patches to chromedp as you want.

PR #290, opened 2019-04-03
Generic code cleanup

kenshaw:
Adding some comments, removing unused items, and renaming handler.go to
target.go to reflect the internal type name changes.


kenshaw:
A new experimental page event was added that I didn't catch before. I updated this cleanup to include the `*page.EventNavigatedWithinDocument` as an ignored event from targets.

Issue #347, opened 2019-05-18
Feature Request: Ability to direct browser stdout/stderr to file


pmurley:
Not following the template because I don't think it's applicable here.

I'd like to be able to capture the browser output (Stdout and Stderr)  into a file of my choosing. Looking briefly at the code in `browser.go`, I think it's probably pretty straightforward to add. I'd be happy to implement this myself and submit as a pull request if you'd consider adding it. If not, I'll just fork and work off of my own version. Let me know your thoughts.

Thanks!

mvdan:
Do you have a specific use case in mind? I've personally never found stdout/stderr useful, besides when something goes really wrong.

Have you double-checked that the devtools protocol doesn't expose those messages somehow? Perhaps through the Runtime domain?

Happy to consider a PR if this is something useful and not possible with our current version.

pmurley:
I am using a custom version of Chromium that prints out data I need to stdout/stderr.

To the best of my understanding, DevTools Protocol exposes console (JavaScript) messages, but not stdout/stderr.

I agree with you that it's probably a rare use case. I'm just trying to avoid forking for as long as possible. Let me try to put together a PR for it and we will see what you think.

mvdan:
Sounds good - we currently ignore all stdout/stderr after we've grabbed the websocket address, so it's unused. Allowing the user to grab it sounds reasonable. This should probably be some sort of ExecAllocator option.

mvdan:
Fixed by #348.

PR #348, opened 2019-05-20
Add options to gather stdout and stderr from browser

pmurley:
Attempt at adding feature in #347.  Exposes both stdout or stderr through an io.Writer. Should not affect users who do not use the added ExecAllocatorOptions `Stdout()` and `Stderr()`.

### DIFF
--- a/allocate.go
+++ b/allocate.go
@@ -164,11 +164,12 @@ func (a *ExecAllocator) Allocate(ctx context.Context, opts ...BrowserOption) (*B
 	}()
 	allocateCmdOptions(cmd)
 
-	// We must start the cmd before calling cmd.Wait, as otherwise the two
-	// can run into a data race.
+	// Create a single reader for both stdout and stderr
 	outputReader, err := cmd.StdoutPipe()
 	cmd.Stderr = cmd.Stdout
 
+	// We must start the cmd before calling cmd.Wait, as otherwise the two
+	// can run into a data race.
 	if err := cmd.Start(); err != nil {
 		return nil, err
 	}
@@ -195,17 +196,23 @@ func (a *ExecAllocator) Allocate(ctx context.Context, opts ...BrowserOption) (*B
 		a.wg.Done()
 		close(c.allocated)
 	}()
-	wsURL, err := addrFromStderr(outputReader)
-	if err != nil {
-		return nil, err
-	}
 
-	// If requested, copy all terminal output from the browser
-	// instance (stdout and stderr) to given io.Writer. Note that
-	// this is different from the output written to the browser console.
+	// If the browser output is requested, use a TeeReader so that we can
+	// grab the websocket address here but also send all output to the
+	// specified writer.
+	var wsURL string
 	if a.combinedOutputWriter != nil {
+		teeReader := io.TeeReader(outputReader, a.combinedOutputWriter)
+		wsURL, err = addrFromStderr(teeReader)
+		if err != nil {
+			return nil, err
+		}
 		go io.Copy(a.combinedOutputWriter, outputReader)
 	} else {
+		wsURL, err = addrFromStderr(outputReader)
+		if err != nil {
+			return nil, err
+		}
 		outputReader.Close()
 	}
 
@@ -227,7 +234,7 @@ func (a *ExecAllocator) Allocate(ctx context.Context, opts ...BrowserOption) (*B
 // addrFromStderr finds the free port that Chrome selected for the debugging
 // protocol. This should be hooked up to a new Chrome process's Stderr pipe
 // right after it is started.
-func addrFromStderr(rc io.ReadCloser) (string, error) {
+func addrFromStderr(rc io.Reader) (string, error) {
 	url := ""
 	scanner := bufio.NewScanner(rc)
 	prefix := "DevTools listening on"
--- a/allocate.go
+++ b/allocate.go
@@ -101,9 +101,6 @@ type ExecAllocator struct {
 
 	wg sync.WaitGroup
 
-	stdout io.ReadCloser
-	stderr io.ReadCloser
-
 	stdoutWriter io.Writer
 	stderrWriter io.Writer
 }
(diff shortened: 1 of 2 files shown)


allocate.go
```diff
@@ -100,6 +100,12 @@ type ExecAllocator struct {
 	initFlags map[string]interface{}
 
 	wg sync.WaitGroup
+
+	stdout io.ReadCloser
+	stderr io.ReadCloser
```
mvdan: seems to me like these don't need to be fields? local vars should suffice

allocate.go
```diff
@@ -363,6 +395,22 @@ func DisableGPU(a *ExecAllocator) {
...
+// will be sent
+func StdoutWriter(writer io.Writer) ExecAllocatorOption {
+	return func(a *ExecAllocator) {
+		a.stdoutWriter = writer
+	}
+}
+
+// StderrWriter is used to set an io.Writer where stderr from the browser
+// will be sent
+func StderrWriter(writer io.Writer) ExecAllocatorOption {
```
mvdan: Are you positive that you need stdout and stderr separately? Not that it's a bad thing, but does chrome even print any stdout separate from stderr?

Also, please name the parameter `w`, not `writer.

allocate.go
```diff
@@ -195,11 +214,25 @@ func (a *ExecAllocator) Allocate(ctx context.Context, opts ...BrowserOption) (*B
...
+			io.Copy(a.stdoutWriter, a.stdout) // Blocks until stderr is closed
+			// Ignore error here because if it fails, we just want to silently
+			// continue and avoid disrupting anything
+		}()
+	}
+
+	if a.stderrWriter != nil {
+		go func() {
+			io.Copy(a.stderrWriter, a.stderr) // Blocks until stderr is closed
+		}()
```
mvdan: do `go f()`, not `go func() { f() }()`.

allocate.go
```diff
@@ -195,11 +214,25 @@ func (a *ExecAllocator) Allocate(ctx context.Context, opts ...BrowserOption) (*B
...
 	if err != nil {
 		return nil, err
 	}
 
+	if a.stdoutWriter != nil {
+		go func() {
+			io.Copy(a.stdoutWriter, a.stdout) // Blocks until stderr is closed
+			// Ignore error here because if it fails, we just want to silently
+			// continue and avoid disrupting anything
+		}()
```
mvdan: same as the other `go func() ...`

allocate.go
```diff
@@ -164,11 +170,24 @@ func (a *ExecAllocator) Allocate(ctx context.Context, opts ...BrowserOption) (*B
...
+	}
+
+	a.stderr, err = cmd.StderrPipe()
+	if err != nil {
+		return nil, err
+	}
+	if a.stderrWriter == nil {
+		defer a.stderr.Close()
+	}
+
```
mvdan: I'd probably move the closes further down, like:

```
if stdoutWriter != nil {
    go io.Copy(...)
} else {
    stdout.Close()
}
```

allocate.go
pmurley: Agreed and done.

allocate.go
pmurley: I'm not totally certain about regular Chrome/Chromium to be honest. I know in my particular use case (custom browser), I do need stdout, but I haven't seen anything from regular Chrome being printed to stdout - it seems to be all stderr.

pmurley:
Implemented all the requested changes, with the exception of combining stdout and stderr.

kenshaw:
May I suggest that someone definitively tests whether or not `stdout` and `stderr` are both used by Chrome before making a high-level API change?

pmurley:
We know it prints most things to stderr. I have an existence proof of the vanilla version of Chromium printing a line ( `Using PPAPI flash.` ) to stdout. [This page](https://developer.chrome.com/native-client/devguide/devcycle/debugging#using-chrome-s-stdout-and-stderr-streams) also seems to confirm that it uses both.

kenshaw:
PPAPI and NaCL have been long long long deprecated.

pmurley:
Absolutely. Just saying that, if the question is "Does regular Chrome/Chromium print stuff to both stdout and stderr?", the answer is yes. Clearly stdout is not used a lot, and as I admitted in #348, I don't think this feature will be widely used, but I need it for my use case, and I could see some other people wanting to grab at least stderr as well.

If you think it's better to combine them into a single output stream, I'm happy to do that - just let me know.

mvdan:
> If you think it's better to combine them into a single output stream, I'm happy to do that - just let me know.

I think that's a good first step. In practically all cases, we'll only have stderr, so users won't be affected. And if anyone wants to capture something extra from stdout, they can also do that.

pmurley:
Alright, made that change. Let me know your thoughts.

allocate.go
```diff
@@ -200,6 +209,17 @@ func (a *ExecAllocator) Allocate(ctx context.Context, opts ...BrowserOption) (*B
 		return nil, err
 	}
 
+	// If requested, copy all terminal output from the browser
+	// instance (stdout and stderr) to given io.Writer. Note that
+	// this is different from the output written to the browser console.
+	if a.browserOutputWriter != nil {
+		go io.Copy(a.browserOutputWriter, stdout)
+		go io.Copy(a.browserOutputWriter, stderr)
```
mvdan: hm, this is bound to be racy. most writers don't accept concurrent writes.

I'd instead use cmd.CombinedPipe, which already takes care of this for you. Our code that picks up the websocket port shouldn't be affected.

allocate.go
```diff
@@ -363,6 +382,14 @@ func DisableGPU(a *ExecAllocator) {
 	Flag("disable-gpu", true)(a)
 }
 
+// BrowserOutput is used to set an io.Writer where stdout and stderr
+// from the browser will be sent
+func BrowserOutput(w io.Writer) ExecAllocatorOption {
```
mvdan: How about CombinedOutput, to be more consistent with os/exec?

allocate.go
```diff
@@ -100,6 +100,8 @@ type ExecAllocator struct {
 	initFlags map[string]interface{}
 
 	wg sync.WaitGroup
+
+	browserOutputWriter io.Writer
```
mvdan: combinedOutputWriter, as per my last comment

allocate.go
pmurley: I don't see any cmd.CombinedPipe() - maybe I'm missing something? There's [cmd.CombinedOutput](https://golang.org/pkg/os/exec/#Cmd.CombinedOutput), but I don't think that's what we want given that it blocks until the command is complete.

pmurley: Using a multireader isn't what we want either, as it would return all of stdout before any stderr (or vice versa). I agree with your concern about races here though.

mvdan: Ah, my bad. It uses Stdout and Stderr directly, not the pipes, so perhaps do the same?

Honestly, we don't care so much about stdout and stderr being in order, since 99% of the time the user will only care about either of those. But two goroutines copying into the writer is definitely buggy.

pmurley:
Thanks for the helpful feedback. Here's another possible solution using a TeeReader so that we gather all output, including the websocket address. Added a test case as well.

mvdan:
I'll squash-merge, since there are many commits in this PR.

mvdan:
@pmurley I've pushed a few commits to master. If you have any thoughts, feel free to send another PR.

PR #361, opened 2019-05-25
fix: AttributesAll possible resource leak

hypocenter:


mvdan:
Sorry, but this isn't enough information. The commit message must explain what the leak is, and why your commit is fixing it.

hypocenter:
'defer' node.RUnlock() calls inside the loop don’t get executed until the func ends , not when each step of for loop ends. All calls here will eat up the func’s stack and may cause unforeseen problems.

mvdan:
You don't have to tell me; you have to specify that in the commit message.

mvdan:
Sorry, I don't know why you closed this. In any case, I'll just commit straight to master.

PR #365, opened 2019-05-28
fix: bug in forceIP function

Pavel-R:
I think this is a bug. The first three characters are separated from the `host` varible (`"ws://localhost:9222/..."` becomes `"alhost:9222/..."`), and further call `net.ResolveIPAddr` returns  an invalid `addr`.


mvdan:
Thanks - you appear to be correct. The tests didn't catch this because they all use IPs or localhost. Not sure why the tests are failing.

Pavel-R:
They shouldn't been fail. According to Travis's log, panic is triggered in tests not related to changes.

PR #372, opened 2019-05-31
add WaitNewTarget

### LINKED ISSUE #371: Consider a helper to easily grab a tab opened by the current tab
We have ExampleClickNewTab, which waits for `EventTargetCreated`. However, this is racy; the target might not have navigated yet, so actions like `Location` might return `about:blank`.

The alternative would be to wait for an `EventTargetInfoChanged` where the URL isn't empty. But perhaps that would never come if the new tab simply remains on `about:blank`?

In any case, this is getting too complex to leave it up to the user. We should probably have a simple helper for the simple use cases.

mvdan:
Means that this common use case can be solved with just five lines of
simple Go code, as opposed to somewhat more complex dozen lines before.

WaitNewTarget also handles the edge cases nicely, such as listening on
both Created and InfoChanged events, ignoring attached targets, and
cancelling the listener once the function matches.

Fixes #371.

### DIFF
--- a/chromedp.go
+++ b/chromedp.go
@@ -443,3 +443,34 @@ func ListenTarget(ctx context.Context, fn func(ev interface{})) {
 		c.targetListeners = append(c.targetListeners, cl)
 	}
 }
+
+// WaitNewTarget can be used to wait for the current target to open a new
+// target. Once fn matches a new unattached target, its target ID is sent via
+// the returned channel.
+func WaitNewTarget(ctx context.Context, fn func(*target.Info) bool) <-chan target.ID {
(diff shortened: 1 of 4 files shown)


chromedp.go
```diff
@@ -443,3 +443,34 @@ func ListenTarget(ctx context.Context, fn func(ev interface{})) {
...
+		default:
+			return
+		}
+		if info.OpenerID == "" {
+			return // not a child target
+		}
+		if info.Attached {
+			return // already attached; not a new target
+		}
+		if fn(info) {
```
rnd: this is neat!

PR #385, opened 2019-06-12
tests: add skip on TestSendKeys/05 for macOS (darwin)

kenshaw:


wubin1989:
I need this feature. When will it be fixed?

PR #389, opened 2019-06-13
actions: Add high-level device emulation action

kenshaw:
Adds two high-level actions, EmulateViewport and ResetViewport that wrap
the Emulation.setDeviceMetricsOverride and Emulation.setTouchEmulationEnabled
commands. Additionally, modifies the two locations where the device
viewport was being changed modified in the unit tests.


kenshaw:
Made the suggested changes.

PR #403, opened 2019-06-26
visibleJS snippet updated

### LINKED ISSUE #402: visibleJS isn't working correctly

I'm trying to click element with fixed position. And `chromedp.Click` freezes forever.

As mentioned here, `offsetParent` can be null even if element is visible.
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent

To determine visibility let's use jQuery implementation
https://github.com/jquery/jquery/blob/e743cbd28553267f955f71ea7248377915613fd9/src/css/hiddenVisibleSelectors.js

So it's just need to change visibleJS to
```js
return !!( elem.offsetWidth || elem.offsetHeight || elem.getClientRects().length );
```

It worked for me, so please, fix it.

ZzEeKkAa:
visibleJS snippet modified to determine value from offsetWidth, offsetHeight and getClientRects().length.

fixes #402

### DIFF
--- a/js.go
+++ b/js.go
@@ -7,14 +7,24 @@ import (
 )
 
 const (
-	// textJS is a javascript snippet that returns the concatenated textContent
-	// of all visible (ie, offsetParent !== null) children.
+	// textJS is a javascript snippet that returns the concatenated innerText of all
+	// visible (ie, offsetWidth || offsetHeight || getClientRects().length ) children.
 	textJS = `(function(a) {
 		var s = '';
 		for (var i = 0; i < a.length; i++) {
-			if (a[i].offsetParent !== null) {
-				s += a[i].textContent;
+			if (a[i].offsetWidth || a[i].offsetHeight || a[i].getClientRects().length) {
+				s += a[i].innerText;
 			}
+		}
+		return s;
+	})(%s)`
+
+	// textContentJS is a javascript snippet that returns the concatenated textContent
+	// of all children.
+	textContentJS = `(function(a) {
+		var s = '';
+		for (var i = 0; i < a.length; i++) {
+			s += a[i].textContent;
 		}
 		return s;
 	})(%s)`
@@ -71,10 +81,11 @@ const (
 		return a[n] = v;
 	})(%s, %q, %q)`
 
-	// visibleJS is a javascript snippet that returns true or false depending
-	// on if the specified node's offsetParent is not null.
+	// visibleJS is a javascript snippet that returns true or false depending on if
+	// the specified node's offsetWidth, offsetHeight or getClientRects().length is
+	// not null.
 	visibleJS = `(function(a) {
-		return a.offsetParent !== null;
+		return Boolean( a.offsetWidth || a.offsetHeight || a.getClientRects().length );
 	})(%s)`
 )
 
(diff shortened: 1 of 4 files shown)


kenshaw:
@ZzEeKkAa appreciate the work here. I'm concerned this won't work in a few edge cases. I had originally written the visible snippet based off code in some other browser driving library, but I don't recall which. Can you give a specific example where `offsetParent` is not null when the element is invisible? Per the MDN, and all the variants I've tried manually in Chrome and using the DevTools inspector, I can't seem to find a case a combination of HTML + CSS that would cause this to be true. If there is a specific case, I honestly would suggest it's a problem on Chrome. I'd be happy to change this specific snippet, but I'd appreciate a pointer to a specific HTML + CSS example where this actually occurs.

kenshaw:
Also, if we merge this, please change the `elem` variable name to `a`, to be consistent with the other JavaScript snippets.

kenshaw:
Sorry; upon closer reading of the MDN, the specific case where this happens is simply where  the element has `position: fixed`. In our testing, this had never been brought up. I had assumed that Chrome would return the `<body>` element, as most other browsers do.

If you change this PR to change the `elem` name, I see no problem in merging this. Also, the `Text()` action has its own `offsetParent` check, which should be changed as well in this PR.

I'd also prefer if the JavaScript was rewritten to not use the `!!`operator.

ZzEeKkAa:
Exactly, with position: fixed. I've left reference to MDN at #403

ZzEeKkAa:
@kenshaw are you sure, that check `"a[i].offsetParent !== null"` in textJS is working properly? In that scope `a[i].offsetParent` is always `undefined`.

ZzEeKkAa:
Found the root of evil: `cashXNode` instead of `cashX` inside `Text`. Changed `textContent` to `innerText` since first one doesn't know about inner hidden elements. Now `Text` function does what it expected to do. See MDN and new tests I've added.

kenshaw:
@ZzEeKkAa appreciate the work. Can you rebase this into a single commit so that it can be reviewed more easily. Also, note that we prefer not having "merge commits", as they don't add any additional information to a repository, and muddy the commit history:

```sh
$ cd ~/src/ && git clone https://github.com/chromedp/chromedp.git
$ cd chromedp
$ git remote add ZzEeKkAa https://github.com/ZzEeKkAa/chromedp.git
$ git fetch ZzEeKkAa && git checkout -b fix-visiblejs-snippet ZzEeKkAa/master
$ git rebase -i origin/master
$ git push -f ZzEeKkAa fix-visiblejs-snippet:master
```

(apologies if the above is incorrect; I have not tried it, and am spitting it from memory)

ZzEeKkAa:
@kenshaw done.

kenshaw:
Thanks. Will review. We're gearing up for a new release, and may not merge this into master until ready with the entire release.

ZzEeKkAa:
Test fails, because text image is not a text. `TestSVGFullXPath` is invalid and has to be modified.

mvdan:
@ZzEeKkAa can you please rebase again and fix any tests as needed to get CI green?

ZzEeKkAa:
Done.

PR #413, opened 2019-07-02
examples: add ExamplePrintToPDF

kenshaw:
Adds a quick example using PrintToPDF, which also show cases using the
cdproto package.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -284,3 +284,27 @@ func ExampleEmulate() {
 		panic(err)
 	}
 }
+
+func ExamplePrintToPDF() {
+	ctx, cancel := chromedp.NewContext(context.Background())
+	defer cancel()
+
+	var buf []byte
+	if err := chromedp.Run(ctx,
+		chromedp.Navigate(`https://godoc.org/github.com/chromedp/chromedp`),
+		chromedp.ActionFunc(func(ctx context.Context) error {
+			var err error
+			buf, _, err = page.PrintToPDF().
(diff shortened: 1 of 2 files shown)


example_test.go
```diff
@@ -284,3 +284,26 @@ func ExampleEmulate() {
...
+	if err := chromedp.Run(ctx,
+		chromedp.Navigate(`https://godoc.org/github.com/chromedp/chromedp`),
+		chromedp.ActionFunc(func(ctx context.Context) error {
+			var err error
+			buf, _, err = page.PrintToPDF().
+				WithDisplayHeaderFooter(false).
+				WithLandscape(true).
+				Do(ctx)
+			return err
+		})); err != nil {
```
mvdan: I prefer to put the call-closing `)` on a separate line, so that all call arguments are consistently on their own lines. That also makes the indentation a bit nicer to follow, I think.

example_test.go
```diff
@@ -284,3 +284,26 @@ func ExampleEmulate() {
 		panic(err)
 	}
 }
+
+func ExamplePrintToPDF() {
+	ctx, cancel := chromedp.NewContext(context.Background(), chromedp.WithDebugf(log.Printf))
```
mvdan: remove the Debugf part? it might trick users into thinking that it's necessary to get the example working, while it isn't.

kenshaw:
@mvdan let me know if that's the change you mean. Appreciated!

PR #418, opened 2019-07-02
target: fix issue with DOM child inserted events

### LINKED ISSUE #412: FullXPath isn't working correctly

<pre>
github.com/chromedp/chromedp v0.0.0-20190617065505-d55cf9043e05
$ google-chrome --version
Version 75.0.3770.100 (Official Build) (64-bit)
$ go version
go version go1.12.1 darwin/amd64
</pre>

Here is an example of page, with dynamic DOM tree, that provides bad xpath:
https://github.com/ZzEeKkAa/chromedp-xpath
Some snippets:
main.go
 ```go
defer cancel()

absPath, err := filepath.Abs("template.html")
if err != nil {
	panic(err)
}
if err := chromedp.Run(ctx, chromedp.Tasks{
	chromedp.Navigate("file://" + absPath),
	chromedp.Query("a.toggle-nav", chromedp.ByQuery),
	chromedp.QueryAfter("a[href='#mm-2']", func(ctx context.Context, nodes ...*cdp.Node) error {
		fullPath := nodes[0].FullXPath()
		fmt.Println("[[FullXPath]]", fullPath)
		if fullPath != "/html[1]/body[1]/div[1]/div[1]/div[1]/ul[1]/li[1]/a[1]" {
			return errors.New("wrong path")
		}
		return nil
	}, chromedp.ByQuery),
}); err != nil {
	log.Fatal(err)
}
```

template.html
```html
<!DOCTYPE html>
<html>
<body>
    <div>
        <a href="#my-menu" class="toggle-nav">Nav</a>
        <div id='my-menu'>
            <ul class="sf-menu">
                <li>
                    <a href="#">Item</a>
                    <ul></ul>
                </li>
            </ul>
        </div>
    </div>
</body>
<script>
    var scr = {
        "scripts": [
            {"src": "./libs.min.js", "async": false},
            {"src": "./script.js", "async": false}
        ]
    };
    !function () {
        "use strict";
        window.onload = function () {
            if ("[object Array]" !== Object.prototype.toString.call(scr.scripts)) return;
            for (var r = 0; r < scr.scripts.length; r++) {
                var c = document.createElement("script"), e = scr.scripts[r];
                c.src = e.src;
                c.async = false;
                document.body.appendChild(c);
            }
        }
    }();
</script>
</html>
```
Unfortunately I don't have sources of libs.min.js. I took it from some site I was playing with.

It's strange, that bug disappears, when scripts are loading through the tag, not through js, or if `chromedp.Query("a.toggle-nav", chromedp.ByQuery),` removed.

xpath to `a[href='#mm-2']` should be `/html[1]/body[1]/div[1]/div[1]/div[1]/ul[1]/li[1]/a[1]`.
You can check it, by open `template.html` in chrome and run 
`$x('/html[1]/body[1]/div[1]/div[1]/div[1]/ul[1]/li[1]/a[1]')[0]` and `document.querySelector('a[href='#mm-2']')` in console.

Instead xpath is `/html[1]/body[1]/div[3]/div[1]/div[1]/ul[1]/li[1]/a[1]`.

kenshaw:
A small issue with the logic of the insertNode nodeOp, and present since
the beginning of the project, meant that element nodes that were copied
from the browser to the Target's Frames/Nodes, that later had a child
element added at the _front_ of the element's children would instead
have the child element node added to the _end_ of the list.

This resulted in incorrect generation of the Node's FullXPath(), and
could cause frustrating/hard to track errors when later attempting to
access that Node using the element query actions.

This change fixes the single one line logic, and adds an
Example_DocumentDump that works as a test for this specific case.
Additionally updates to the latest cdproto in order to make use of
Node.Dump in the example test case.

Note: the Example_DocumentDump test may be flaky at the moment, as it
includes the NodeIDs of the dumped DOM elements, which can vary from
Chrome runs. This is planned to be fixed before the next release.

Fixes #412.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -345,3 +345,60 @@ func ExampleByJSPath_OuterHTML() {
 	//	<div id="content">cool content</div>
 	//</body></html>
 }
+
+func Example_DocumentDump() {
+	ctx, cancel := chromedp.NewContext(context.Background())
+	defer cancel()
+
+	ts := httptest.NewServer(writeHTML(`<!doctype html>
+<html>
+<body>
+  <div id="content">the content</div>
+</body>
+</html>`))
+	defer ts.Close()
+
+	const expr = `(function(d, id, v) {
+		var b = d.querySelector('body');
+		var el = d.createElement('div');
+		el.id = id;
+		el.innerText = v;
+		b.insertBefore(el, b.childNodes[0]);
+	})(document, %q, %q);`
+
+	var nodes []*cdp.Node
+	if err := chromedp.Run(ctx,
+		chromedp.Navigate(ts.URL),
+		chromedp.Nodes(`document`, &nodes, chromedp.ByJSPath),
+		chromedp.WaitVisible(`#content`),
+		chromedp.ActionFunc(func(ctx context.Context) error {
+			s := fmt.Sprintf(expr, "thing", "a new thing!")
+			_, exp, err := runtime.Evaluate(s).Do(ctx)
+			if err != nil {
+				return err
+			}
+			if exp != nil {
+				return exp
+			}
+			return nil
+		}),
+		chromedp.WaitVisible(`#thing`),
+	); err != nil {
+		panic(err)
+	}
+
+	fmt.Println("Document tree:")
+	fmt.Print(nodes[0].Dump("  ", "  ", false))
+
+	// Output:
+	// Document tree:
+	//   #document <Document>
+	//     html <DocumentType>
+	//     html
+	//       head
+	//       body
+	//         div#thing
+	//           #text "a new thing!"
+	//         div#content
+	//           #text "the content"
+}
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/chromedp/chromedp
 go 1.11
 
 require (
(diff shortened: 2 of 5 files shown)


kenshaw:
@mvdan let's hold off on merging this, until we've been able to better verify if this is the correct logic. Also, I plan on fixing the example test case / Node.Dump so that it isn't flaky.

mvdan:
Sounds good. CI is red, by the way.

kenshaw:
@mvdan I've fixed the flaky unit test, as well as pushed a new unit test that explicitly tests the following `nodeOp`'s:

`childNodeInserted`
`childNodeRemoved`

example_test.go
```diff
@@ -345,3 +345,60 @@ func ExampleByJSPath_OuterHTML() {
...
+		el.id = id;
+		el.innerText = v;
+		b.insertBefore(el, b.childNodes[0]);
+	})(document, %q, %q);`
+
+	var nodes []*cdp.Node
+	if err := chromedp.Run(ctx,
+		chromedp.Navigate(ts.URL),
+		chromedp.Nodes(`document`, &nodes, chromedp.ByJSPath),
+		chromedp.WaitVisible(`#content`),
```
azbshiri: Didn't know there's a `WaitVisible`, great!

PR #451, opened 2019-08-14
Minor cleanup

kenshaw:
Cleans up .gitignore, and adds --rm to the docker run in travis-ci.


kenshaw:
@mvdan I'm adding this so that when I copy/paste from the travis config, it automatically removes the container. Also makes sense because this is a "transient" container, and not really useful to keep it kicking around on disk.

PR #460, opened 2019-08-23
Websocket URL read timeout provided.

ZzEeKkAa:
In some cases it is impossible to parse websocket url from chrome's stdout. This PR provides way to prevent program stuck.

### DIFF
--- a/allocate.go
+++ b/allocate.go
@@ -3,12 +3,14 @@ package chromedp
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
 	"sync"
+	"time"
 )
 
 // An Allocator is responsible for creating and managing a number of browsers.
@@ -47,6 +49,7 @@ func setupExecAllocator(opts ...ExecAllocatorOption) *ExecAllocator {
 // if the given parent context doesn't have an allocator set up. Do not modify
 // this global; instead, use NewExecAllocator. See ExampleExecAllocator.
 var DefaultExecAllocatorOptions = [...]ExecAllocatorOption{
+	WSURLReadTimeout(time.Second * 10),
 	NoFirstRun,
 	NoDefaultBrowserCheck,
 	Headless,
@@ -101,6 +104,7 @@ type ExecAllocator struct {
 
 	wg sync.WaitGroup
 
+	wsURLReadTimeout     time.Duration
 	combinedOutputWriter io.Writer
 }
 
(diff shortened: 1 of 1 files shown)


mvdan:
What kind of scenario did you encounter this with? I've only seen chrome either crashing, or giving the URL. If there's a way for it to hang reliably without giving the URL, we should add a test for it.

ZzEeKkAa:
It happens randomly. During the day it can happen about 10-100 times in 360000 calls approx. I could not figure out what was the root of the problem. Just see goroutines stuck at time of reading websocket url. With this pr problem gone. Don't have much time to investigate this issue in the near future.

mvdan:
Thanks. I'm going to leave this open for now; I'd rather not fix bugs that cannot be tested or reproduced. Perhaps others have input in the future.

kenshaw:
For what it's worth, I'm very aware of this issue, and I've encountered it a number of times. I'm looking into other remedies beyond this, however I am making a point of fixing this issue permanently.

allocate.go
```diff
@@ -202,7 +206,23 @@ func (a *ExecAllocator) Allocate(ctx context.Context, opts ...BrowserOption) (*B
 		close(c.allocated)
 	}()
 
-	wsURL, err := readOutput(stdout, a.combinedOutputWriter, a.wg.Done)
+	var wsURL string
+	if a.wsURLReadTimeout > 0 {
+		wsURLChan := make(chan struct{})
```
mvdan: make this buffered with cap==1

allocate.go
```diff
@@ -202,7 +206,23 @@ func (a *ExecAllocator) Allocate(ctx context.Context, opts ...BrowserOption) (*B
...
+		go func() {
+			wsURL, err = readOutput(stdout, a.combinedOutputWriter, a.wg.Done)
+			wsURLChan <- struct{}{}
+		}()
+		select {
+		case <-wsURLChan:
+		case <-time.After(a.wsURLReadTimeout):
+			err = errors.New("websocket url timeout reached")
+			// preventing goroutine leak.
+			go func() { <-wsURLChan }()
```
mvdan: you can remove this if you make the channel buffered, as explained above

allocate.go
```diff
@@ -401,6 +421,14 @@ func CombinedOutput(w io.Writer) ExecAllocatorOption {
 	}
 }
 
+// WSURLReadTimeout is used to set a timeout for reading websocket url on
+// chrome process starting.
+func WSURLReadTimeout(timeout time.Duration) ExecAllocatorOption {
```
mvdan: I don't think we want to make this configurable for now. A constant of 10s should be plenty for any machine to start Chrome, yet still be short enough that we don't hang a process for a long time.

Make the 10s a constant instead of a field, too.

mvdan:
I've merged this and done the changes above myself; hope that's okay.

PR #479, opened 2019-10-05
fix bug,time.After will never execute.

lovelly:
time.After will never execute.


lovelly:
This should be caused by another reason, I closed it.

PR #482, opened 2019-10-09
flatten version 2

mvdan:


mvdan:
This was posted just to show an alternative approach.

PR #483, opened 2019-10-09
Update allocate.go

lchjczw:
add chrome.exe path

### DIFF
--- a/allocate.go
+++ b/allocate.go
@@ -319,7 +319,7 @@ func findExecPath() string {
 		"chrome",
 		"chrome.exe", // in case PATHEXT is misconfigured
 		`C:\Program Files (x86)\Google\Chrome\Application\chrome.exe`,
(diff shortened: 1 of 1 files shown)


kenshaw:
We appreciate the contribution, but this is a "non-standard" location for the Chrome executable. Understand that it is not tenable in `findExecPath` to list every possible location that the executable might be installed at.

That said, the "correct" way to set this for use in your application is to use the [`ExecPath` allocator option](https://godoc.org/github.com/chromedp/chromedp#ExecPath), which will force the path used the executable.

lchjczw:
I don't known  how to get the standard location ,And whether  the  chrome is same in the global,  but this is default install location for the Chrome executable in china.
Thank you for your reminder. 
In addition to the above method, I used to add the chrome.exe path to the global variable.

PR #487, opened 2019-10-13
Fix concurrency issues

kenshaw:
Intermittent timeouts on certain channel reads can occur, and not catch
when a context has been closed causing hangs and other potential issues.

This adds additional checks for when a context has been closed in a
first pass at fixing the fundamental issue.

Additionally, gets rid of the `raw{Marshal,Unmarshal}` calls, and
updates LICENSE.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2016-2018 Kenneth Shaw
+Copyright (c) 2016-2020 Kenneth Shaw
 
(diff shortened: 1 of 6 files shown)


molon:
@kenshaw 
I have encountered problems in this place.
https://github.com/kenshaw/chromedp/blob/a76972444aa378c1ad34715e64b308ee19faadad/chromedp.go#L274

kenshaw:
@molon apologies. I had missed these other instances on fixing the context read. I'll be pushing an update shortly covering these instances.

PR #494, opened 2019-10-19
Enable remote-debugging-port parameter

aaron0x:
If this parameter is set, don't force it to be 0.

### DIFF
--- a/allocate.go
+++ b/allocate.go
@@ -148,7 +148,9 @@ func (a *ExecAllocator) Allocate(ctx context.Context, opts ...BrowserOption) (*B
 		// default, unless the user set Flag("no-sandbox", false).
 		args = append(args, "--no-sandbox")
 	}
-	args = append(args, "--remote-debugging-port=0")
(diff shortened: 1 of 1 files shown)


aaron0x:
When user also sets remote-debugging-port, there are two remote-debugging-port passed to chrome, and chrome only take the later one, that's 0. debugging port can never be set.

allocate.go
```diff
@@ -377,6 +379,11 @@ func UserAgent(userAgent string) ExecAllocatorOption {
 	return Flag("user-agent", userAgent)
 }
 
+// RemoteDebuggingPort is the command line option to set remote-debugging-port.
+func RemoteDebuggingPort(port string) ExecAllocatorOption {
```
mvdan: Remove this API. Most users shouldn't be setting the parameter at all, and those who know what they're doing can just call Flag directly.

aaron0x: ok

mvdan:
Thanks. The builds are broken in master; that's unrelated to your change, which seems pretty safe and sensible to me.

PR #497, opened 2019-10-25
Added .idea to .gitignore for goland IDE

folkengine:


mvdan:
Sorry, but no. Set up a global gitignore file instead of adding this to every single project.

PR #506, opened 2019-11-03
Update dependencies and go version

sukso96100:


mvdan:
The failures are because of a newer chrome version breaking our tests. We should pin a specific version; I'll take care of that.

PR #511, opened 2019-11-17
docs: fix broken links to devices list

cstrouse:
Fixes the broken links in the inline documentation for the list of off-the-shelf device types.

### DIFF
--- a/emulate.go
+++ b/emulate.go
@@ -82,7 +82,7 @@ func ResetViewport() EmulateAction {
 
 // Device is the shared interface for known device types.
 //
-// See: github.com/chromedp/chromedp/device for a set of off-the-shelf devices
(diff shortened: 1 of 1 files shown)


kenshaw:
It's worth noting that the `See: ` documentation here is the _package_ name. If you wanted this to be a URL, I won't consider merging the GitHub URL, as it's not very helpful. If you want to change this with the godoc.org URL for this package path, I'll think about including this, however I'd like to point out that from a documentation standpoint this is wildly inconsistent and would be non-idiomatic. I'm all for making the documentation more helpful, especially to new users of the package, but I don't know if a pointer to the source code accomplishes that. At least with the GoDoc URL, users would be able to click on the the link in a terminal and open a browser or some such.

Issue #472, opened 2019-09-18
Have Cancel gracefully close the Browser


pmurley:
I am doing some work collecting code coverage data on Chromium (not JavaScript coverage, but coverage of the browser source/C++). In order for Chromium to correctly generate the output files containing this coverage data, I need it to close gracefully. I believe the DevTools Protocol `Browser.close()` would work fine.

1. My understanding is that when you cancel a context, the associated browser is closed with SIGKILL. Is that accurate?

2. If so, is there an easy way I can circumvent that to close the browser gracefully? I think that right now, [this if statement](https://github.com/chromedp/chromedp/blob/ff5b4fce079ec7e7e53fc16fd9f1b47569e9ed4b/browser.go#L166) is preventing it. Will there be unintended effects if I just remove that check?

mvdan:
Interesting; I didn't know there were use cases where gracefully closing down the entire browser was necessary. I assume gracefully closing down tabs (which is already handled) doesn't cut it for you?

The graceful closing of tabs doesn't happen for the original tab, because the browser can't be left with no tabs.

Your assumptions are correct, I think. We leave `os/exec` handle stopping the process on context cancellation, and it uses `os.Process.Kill`, which seems to just send `SIGKILL`.

I don't think we want `defer cancel()` to do this, for two reasons. First, most users won't need graceful shutdown, which could take extra time or give uninteresting errors. Second, the `cancel()` func alone can't provide any error, while a graceful shutdown needs to report an error; otherwise it's not graceful at all.

So I think the best option is to add this to the `chromedp.Cancel` API. It would send `Browser.close` to the browser, wait for it to report it's finished, and then cancel the underlying context and clean up any leftover resources. The context argument could be wrapped with a timeout too, if one wants to limit the amount of time given to Chrome to close gracefully:

```
// Give Chrome five seconds to shut down properly. After that, kill it.
chromedp.Cancel(context.WithTimeout(ctx, 5 * time.Second))
```

I'd still forbid direct use of `Browser.close` via `chromedp.Run`. If it's available via `chromedp.Cancel`, it doesn't make sense to also allow it via `Run`. We'd need extra code and tests to ensure it does the right thing and doesn't hang chromedp, and I just don't see the point :)

pmurley:
Understood -- I think the solution you propose would satisfy my use case. Unfortunately, I'm not able to submit a pull request right now, so no problem if it's not a priority. Thanks as always!

pmurley:
[...]

PR #539, opened 2019-12-10
fix to gracefully close browser

alanrothschild:
Modified chromedp.Cancel method to attempt to gracefully close browser instead of relying on os.Process.Kill that sends SIGKILL

Fixes "Chrome didn't shut down correctly" message on subsequent launch of browser after it is killed.

Code is borrowed from snippet posted by pmurley in question #472


mvdan:
The fix with the test heavily inspired by your code was merged a few days ago, so I'll assume that you're happy with that. Closing - thanks!

PR #556, opened 2020-01-15
Remove panic that could be error

nicolascb:
Removes panic in some situations where error could be returned.

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -3,6 +3,7 @@ package chromedp
 import (
 	"context"
 	"encoding/json"
+	"errors"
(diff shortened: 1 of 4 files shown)


mvdan:
Is there a reason you're removing panics in each one of these cases?

Most of the panics you are removing are intended to make the library crash when you don't follow the documentation. For example, passing a nil pointer to a function that specifically says it only accepts non-nil pointers.

That's a correct use of panic. If you're encountering panics in production, you should be testing your code a bit better :)

Issue #555, opened 2020-01-14
Optionally avoid actions in `chromedp.attachTarget()`


DCtheTall:
#### What versions are you running?

Chrome: 79.0.3945.117

#### What is the issue?

When you start a new browser context or attach to a new target, then by calling `chromedp.Run` you implicitly call a number of actions:

https://github.com/chromedp/chromedp/blob/master/chromedp.go#L316

It would be nice if it were possible to optionally avoid calling these actions on a new target. For example, if I want the debugger to attach to a web worker, then the method `Page.enable` is not available and the first call of `chromedp.Run` will fail, but subsequent calls will not.

I'd be interesting in helping with a PR if this is a change people support and there is consensus on what the user-facing interface for this would be.

mvdan:
I'm not sure I follow. If you're talking to something that implements only part of the devtools protocol, is chromedp really useful? For example, the main purpose of the library is to manage targets, and thus pages.

Some other actions we absolutely depend on. For example, discovering targets and using the flattened protocol.

I'm really not sure how we could make this "initial actions" stage more powerful. Allowing the user to override these entirely or one-by-one would just be confusing for the vast majority of users, who just want to attach to a Chrome browser.

DCtheTall:
Not to be rude, but that is kind of a disappointing answer :(

I understand that users depend on some of these CDP methods in their code, and many people might find it confusing if these methods weren't called before they started doing their own work on the page.

On the other hand, as a user of the library, I find the library internally calling many CDP methods is confusing. It is not unheard of for people to use the CDP to inspect worker targets where the Page.* protocol is not implemented. It is a tricky implementation detail imho that you need to let `chromedp.Run` fail once in order to use it in a worker. Perhaps this is more of a documentation issue?

mvdan:
We seem to be talking past each other. I just don't understand how chromedp is useful to you right now, if the Page domain is entirely unavailable. Otherwise, we're talking about a solution before [the problem is clear](https://en.wikipedia.org/wiki/XY_problem).

Do you have links about worker targets? Or do you mean [service workers](https://developers.google.com/web/fundamentals/primers/service-workers)? Google search results aren't being very clear.

I'm all for making the library more useful or powerful, but I also want to keep the API small where possible. In this case, I want to fully understand what the use case is before I go ahead with exposing internal details to the API.

DCtheTall:
Thank you for explaining and sorry for my own misinterpretation.

[...]

PR #565, opened 2020-01-27
Allow chromedp to attach to worker targets without error (#555)

DCtheTall:
The way we tell if a new target is a web worker or a service worker is by using `Runtime.evaluate` and the fact that `self` in workers will refer to either a `[Service]WorkerGlobalScope`.

Fixes [issue #555](https://github.com/chromedp/chromedp/issues/555)

### DIFF
--- a/chromedp.go
+++ b/chromedp.go
@@ -12,6 +12,7 @@ package chromedp
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
@@ -247,12 +248,14 @@ func (c *Context) newTarget(ctx context.Context) error {
 		// This new page might have already loaded its top-level frame
 		// already, in which case we wouldn't see the frameNavigated and
 		// documentUpdated events. Load them here.
-		tree, err := page.GetFrameTree().Do(cdp.WithExecutor(ctx, c.Target))
-		if err != nil {
-			return err
+		if !c.Target.isWorker {
+			tree, err := page.GetFrameTree().Do(cdp.WithExecutor(ctx, c.Target))
+			if err != nil {
+				return err
+			}
+			c.Target.cur = tree.Frame
+			c.Target.documentUpdated(ctx)
 		}
-		c.Target.cur = tree.Frame
-		c.Target.documentUpdated(ctx)
 		return nil
 	}
 	if !c.first {
@@ -313,19 +316,33 @@ func (c *Context) attachTarget(ctx context.Context, targetID target.ID) error {
 	c.Target.listeners = append(c.Target.listeners, c.targetListeners...)
 	go c.Target.run(ctx)
 
-	for _, action := range []Action{
-		// enable domains
+	if err := runtime.Enable().Do(cdp.WithExecutor(ctx, c.Target)); err != nil {
+		return err
+	}
+
+	res, _, err := runtime.Evaluate("self;").Do(cdp.WithExecutor(ctx, c.Target))
+	if err != nil {
+		return err
+	}
+	c.Target.isWorker = strings.Contains(res.ClassName, "WorkerGlobalScope")
+
+	// Enable available domains and discover targets.
+	actions := []Action{
 		log.Enable(),
 		runtime.Enable(),
-		inspector.Enable(),
-		page.Enable(),
-		dom.Enable(),
-		css.Enable(),
+	}
+	if !c.Target.isWorker {
+		actions = append(actions, []Action{
+			inspector.Enable(),
+			page.Enable(),
+			dom.Enable(),
+			css.Enable(),
+			target.SetDiscoverTargets(true),
+			target.SetAutoAttach(true, false).WithFlatten(true),
+		}...)
+	}
 
-		// enable target discovery
-		target.SetDiscoverTargets(true),
-		target.SetAutoAttach(true, false).WithFlatten(true),
-	} {
+	for _, action := range actions {
 		if err := action.Do(cdp.WithExecutor(ctx, c.Target)); err != nil {
 			return fmt.Errorf("unable to execute %T: %v", action, err)
 		}
--- a/chromedp.go
+++ b/chromedp.go
@@ -316,10 +316,12 @@ func (c *Context) attachTarget(ctx context.Context, targetID target.ID) error {
 	c.Target.listeners = append(c.Target.listeners, c.targetListeners...)
 	go c.Target.run(ctx)
 
+	// Check if this is a worker target. We cannot use Target.getTargetInfo or
+	// Target.getTargets in a worker, so we check if `self` refers to a
+	// WorkerGlobalScope or ServiceWorkerGlobalScope.
 	if err := runtime.Enable().Do(cdp.WithExecutor(ctx, c.Target)); err != nil {
 		return err
 	}
-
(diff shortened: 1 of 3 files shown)


mvdan:
Hm. This adds another step for every new target, which I'd like to avoid if we can. How about something like:

```
for _, action := range actions {
	if err := action.Do(cdp.WithExecutor(ctx, c.Target)); isNotImplemented(err) {
		// The domain is not available; this could be a web worker.
	} else if err != nil {
		return fmt.Errorf("unable to execute %T: %v", action, err)
	}
}
```

`isNotImplemented` would ideally use the type system to check that the error is the particular one we want, but given that the errors come from Chrome, I presume you'll have to do a substring match for now.

Another advantage of this method is that it only skips the domains that aren't implemented, instead of all of them at once given some previous check. What if a `devtools` target supports `page` but not `inspector`? Not that I can think of a specific case right now, but it doesn't seem impossible.

DCtheTall:
@mvdan I wanted to do that originally, but there are a two things that led me to think this design might be preferable:

1. Not all the errors are of the same form. For instance, `[inspector|page|dom|css].enable` cause it to throw a "not available" error (code is -32601), whereas `Target.setAutoAttach` and `Target.setDiscoverTargets` throw a "not allowed" error (code -32000). We could cover both error types with string matching but I think the method I chose is a bit more sturdy.

2. `chromedp.newTarget()` also tries to call `Page.getFrameTree` on all new targets, which are not available on workers, since the `Page` domain is not implemented. So we need some way of keeping a record of whether the new target is a worker to avoid this. Again, I think the method in this PR is preferable to string matching, even if it is an extra command.

My code in particular just checks if the target is a worker and avoids what is not implemented in workers. I think that it is reasonable to address other targets that do not fully implement CDP on a case-by-case basis. I agree it's not impossible but I also think it isn't reasonable to try to predict what combinations are allowed/not allowed.

Feel free to disagree with anything I said here!

chromedp.go
```diff
@@ -313,19 +316,35 @@ func (c *Context) attachTarget(ctx context.Context, targetID target.ID) error {
...
+	res, _, err := runtime.Evaluate("self;").Do(cdp.WithExecutor(ctx, c.Target))
+	if err != nil {
+		return err
+	}
+	c.Target.isWorker = strings.Contains(res.ClassName, "WorkerGlobalScope")
+
+	// Enable available domains and discover targets.
+	actions := []Action{
 		log.Enable(),
 		runtime.Enable(),
```
mvdan: this is now duplicate, no?

chromedp.go
```diff
@@ -313,19 +316,35 @@ func (c *Context) attachTarget(ctx context.Context, targetID target.ID) error {
...
-		dom.Enable(),
-		css.Enable(),
-
-		// enable target discovery
-		target.SetDiscoverTargets(true),
-		target.SetAutoAttach(true, false).WithFlatten(true),
-	} {
+	}
+	if !c.Target.isWorker {
+		actions = append(actions, []Action{
```
mvdan: add a short comment about how these domains aren't supported on workers (ideally with a link to some docs)

chromedp_test.go
```diff
@@ -716,3 +718,64 @@ func TestGracefulBrowserShutdown(t *testing.T) {
...
+	mux := http.NewServeMux()
+	mux.Handle("/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "text/html")
+		io.WriteString(w, `<html>
+      <body>
+        <script>
+          new Worker('/worker.js');
+        </script>
+      </body>
+    </html>`)
```
mvdan: use tabs here

chromedp_test.go
```diff
@@ -716,3 +718,64 @@ func TestGracefulBrowserShutdown(t *testing.T) {
...
+	defer cancel()
+
+	ch := make(chan error, 1)
+
+	ListenTarget(ctx, func(ev interface{}) {
+		if ev, ok := ev.(*target.EventAttachedToTarget); ok {
+			if ev.TargetInfo.Type != "worker" {
+				return
+			}
+			go func() {
```
mvdan: hm, when I suggested the use of a channel, I was thinking more like:

```
ListenTarget(...) {
    ch <- someValue
}
Run(Navigate(...))
value := <-ch
Run(...)
```

That is, keep all the actions in the main goroutine, to keep the flow of actions linear. You also don't need the extra goroutine, since your channel is buffered.

mvdan:
You're right about newTarget, I hadn't noticed that.

I think you're also right about overthinking future use cases other than workers. If they come by, we can always add support for them later, given that this is not any new API.

DCtheTall:
FWIW I noticed that `TestGracefulBrowserShutdown` seemed to be flaky, I am not sure if you knew this already.

mvdan:
Hm. Yeah, I've noticed that a couple of times, but haven't investigated yet. It's a pretty new test, but I couldn't reproduce any failures locally.

chromedp.go
```diff
@@ -313,19 +316,35 @@ func (c *Context) attachTarget(ctx context.Context, targetID target.ID) error {
 	c.Target.listeners = append(c.Target.listeners, c.targetListeners...)
 	go c.Target.run(ctx)
 
-	for _, action := range []Action{
-		// enable domains
+	// Check if this is a worker target. We cannot use Target.getTargetInfo or
+	// Target.getTargets in a worker, so we check if `self` refers to a
```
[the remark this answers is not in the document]
DCtheTall: Done.

chromedp.go
DCtheTall: It is, removed.

chromedp.go
DCtheTall: Comment added. I am unaware of any documentation of that explicitly states that the `Page` domain is not supported in web workers. I can ask around, but as you probably know the Chrome DevTools Protocol is sometimes not well documented.

chromedp_test.go
DCtheTall: Done.

chromedp_test.go
DCtheTall: Done.

DCtheTall:
Do you think it is worth adding a test case for service workers as well? My code should work for both types of workers, but maybe adding test coverage will help keep it working?

mvdan:
Yeah, I would say so. If you can reuse much of the test code, then I'd suggest using subtests, like `TestWithWorker/Web` and `TestWithWorker/Service`, or `TestNonBrowesr/WebWorker` etc.

DCtheTall:
@mvdan SGTM. I will make test cases for the three types of workers (service, dedicated, and shared). The test for each case is essentially the same so it should be possible to test all 3 under the same test case using subtests. I'll tag you in another comment when that is ready for review.

DCtheTall:
@mvdan Added a test case for service workers. I could not get a test using a SharedWorker to work. I am admittedly unfamiliar with how SharedWorkers interact with the DevTools, so I pushed what I have so far.

The failing check is due to the flake I mentioned in a previous comment.

DCtheTall:
@mvdan thanks for the review, merge away!

mvdan:
It appears I got distracted :) I'll do it when I'm on the keyboard again.

PR #566, opened 2020-01-27
Add GetBrowserPid option to browser.go

Ajod:
The process PID can come in handy when there is talk about running a third party software against a remotely controlled browser.

New PR from clean fork

### DIFF
--- a/browser.go
+++ b/browser.go
@@ -332,3 +332,19 @@ func WithConsolef(f func(string, ...interface{})) BrowserOption {
 func WithDialTimeout(d time.Duration) BrowserOption {
 	return func(b *Browser) { b.dialTimeout = d }
 }
+
+// GetBrowserPid returns the process ID of the browser attached to the provided context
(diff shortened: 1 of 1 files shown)


mvdan:
> The process PID can come in handy when there is talk about running a third party software against a remotely controlled browser.

I don't get what the purpose here is. Can you provide an example?

Also, the API seems pretty stiff. Why not expose https://golang.org/pkg/os/#Process? We could expose and document the field in `Browser`. It wasn't done initially because, in general, it's not very useful.

mvdan:
Friendly ping @Ajod. This would also need a test before merging.

Ajod:
Hi, based on your comment that it might not have a purpose I eventually ended up not needing to expose the PID or the Process at all. This would probably a somewhat pointless PR and I should probably close it if that is okay with you

mvdan:
Sounds good, thanks!

aohanhongzhi:
hi guys. I want to kown how to get the browser pid when I open it by chromedp.

Issue #440, opened 2019-07-24
chromedp.WaitVisible("#foot") hangs but WaitVisible("foot", chromedp.ByID) works


kjk:
Go: `go version go1.12.7 windows/amd64`

chromedp:
```
github.com/chromedp/cdproto v0.0.0-20190721111337-61a0348ea0b1
github.com/chromedp/chromedp v0.3.1
```

Full test program: https://gist.github.com/kjk/804bca2d2dfc102f73d8817d4997d172

The core:
```go
qe := url.QueryEscape(searchTerm)
uri := fmt.Sprintf("https://www.google.com/search?q=" + qe)
fmt.Printf("before run, uri='%s'\n", uri)
sel := "div.rc > div:first-child > a:first-child"
var nodes []*cdp.Node
err := chromedp.Run(ctx,
	chromedp.Navigate(uri),
	chromedp.WaitVisible("#foot"),
	chromedp.Nodes(sel, &nodes),
)
```
This hangs in `WaitVisible("#foot").
It works when I replace:
```go
chromedp.WaitVisible("#foot"),
```
with:
```go
chromedp.WaitVisible(`foot`, chromedp.ByID),
```

Those should be equivalent but it looks like `BySearch` version doesn't work.

I also verified in dev tools console that `$("#foot")` works (which I assume uses the same browser API as `WaitVisible("#foot")` would end up using).

mvdan:
Thanks for the detailed report. Did you test with master by any chance?

kjk:
As far as I can tell, it also happens on master.

mvdan:
Thanks. I typed the following reply three days ago, but I think it got lost in the ether due to the poor connection I had at the time.

The misconception here is that the default `By` func isn't `ByQuery`, but `BySearch` - which is pretty much a plaintext search over the page's HTML. And because the `#foo` string appears three times, and not all of those occurrences are contained by visible nodes, the `Wait` function hangs forever.

Try `chromedp.WaitVisible("#foot", chromedp.ByQuery)` - I'm willing to bet it will work, because `ByID` is simply that, plus ensuring the leading `#`.

The default is documented, as you can see on the godoc:

>  By* options specify the type of element query used By the browser to perform the selection query. When not specified, element queries will use BySearch (a wrapper for DOM.performSearch). 

However, that perhaps isn't very obvious or clear. If you have a suggestion to improve the docs or add an entry to the FAQ, I'll gladly accept a PR.

kjk:
[...]

PR #590, opened 2020-03-27
Document that BySearch matches by plain text too

rhcarvalho:
Related to #440.


mvdan:
This is from https://chromedevtools.github.io/devtools-protocol/tot/DOM#method-performSearch:

> Plain text or query selector or XPath search query.

So I agree - the wording seems to match what you've updated it to.

PR #609, opened 2020-05-01
fix: docs for ListenTraget and ListenBrowser were swapped

marco-m:
I _think_ that the sentence

> Note that this only includes browser events; command responses and target events are not included.

was incorrectly added to the documentation for `ListenTarget`, while it refers to `ListenBrowser`.

If I am wrong, then the names for these two functions are a bit confusing :-)

* Function ListenTarget has been added in https://github.com/chromedp/chromedp/commit/39c308a94a7d95a4297791821f6c0df6edf7af52
* Function ListenBrowser has been added in https://github.com/chromedp/chromedp/commit/50c80f91be4ddefee7e881df4613d3ac3fc1aa3a

### DIFF
--- a/chromedp.go
+++ b/chromedp.go
@@ -455,8 +455,9 @@ type cancelableListener struct {
 }
 
 // ListenBrowser adds a function which will be called whenever a browser event
-// is received on the chromedp context. Cancelling ctx stops the listener from
-// receiving any more events.
+// is received on the chromedp context. Note that this only includes browser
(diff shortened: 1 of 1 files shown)


marco-m:
The two failing CI checks fail with `We are currently unable to download the log. Please try again later.`, so I cannot understand what went wrong. This PR only changes two comments.

mvdan:
Whoops, thanks for noticing :) I was probably jumping between the two functions and forgot where I was.

PR #619, opened 2020-05-09
Use %ProgramFiles% and %LocalAppData% to locate chrome.exe more critically

szzhiyang:
Because `Program Files` folder and `Program Files (x86)` folder are not always under C: drive, 
and `Local` folder is not always under `%UserProfile%\AppData`.

### DIFF
--- a/allocate.go
+++ b/allocate.go
@@ -336,10 +336,10 @@ func findExecPath() string {
 
 		// Windows
 		"chrome",
-		"chrome.exe", // in case PATHEXT is misconfigured
-		`C:\Program Files (x86)\Google\Chrome\Application\chrome.exe`,
-		`C:\Program Files\Google\Chrome\Application\chrome.exe`,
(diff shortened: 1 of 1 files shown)


szzhiyang:
@mvdan 

Hi, I've added fall-back paths with `C:\` and `%UserProfile%`.

According to [documentation](https://golang.org/pkg/path/filepath/#Join), the result of `filepath.Join` is always [**Cleaned**](https://golang.org/pkg/path/filepath/#Clean), which does not contain extra slash on the right.

mvdan:
Sorry, but I didn't mean duplicating the entries. For each one of those, you check the var version if the var is defined, otherwise you check the older hard-coded path we have in master.

mvdan:
Also, I know that filepath.Join cleans. But `Join(foo, "bar")` is still simpler than `Join(foo, "/bar")`

ZekeLu:
In my opinion, `findExecPath` makes a best-effort to find the browser. It does not have to cover all cases. The users can always use `chromedp.ExecPath` to specify the path if `findExecPath` does not work. Closing.

PR #622, opened 2020-05-13
add a QueryOption to run a query from a specified node

### LINKED ISSUE #463: support ByQuery and ByQueryAll on specified nodes
like puppeteer

```
	let li = await this.page.$$(`.s-result-item`)
	for (var y = 0; y < li.length; y++) {
		let itemAsin = await this.page.evaluate(element => element.getAttribute('data-asin'), li[y])
		if(itemDATA == posData) {// hit data
			hit = true;
			await li[y].$eval('a.s-links', e => e.click())
			return
		}
	}
```

maybe like

```
let eles = document.querySelector('.s-result-item')

eles[0].querySelector('a.s-links')
```

this is my code

```
chromedp.Run(c.ctx,
			chromedp.WaitVisible(`.a-pagination`, chromedp.ByQuery),
			chromedp.Nodes(`.s-result-item`, &nodes, chromedp.ByQueryAll),
		)
		for _, item := range nodes {
			cur := item.AttributeValue("data-id")
			if cur == searchID {
				fmt.Printf("hit %s = %s\n", cur, searchID)
				chromedp.Run(c.ctx, chromedp.Nodes(item.PartialXPathByID()+` //h2/a[@class="a-text-normal"]`, chromedp.BySearch)) // invalid 
	
				c.Sleep(2 * time.Second)
				return nil
			}
		}
```

mvdan:
Called FromNode, it means that we can run a query rooted at a specific
node, instead of the default which is the root node of the document.

Fixes #463.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -398,6 +398,56 @@ func ExampleByJSPath() {
 	// </body></html>
 }
 
+func ExampleFromNode() {
+	ctx, cancel := chromedp.NewContext(context.Background())
+	defer cancel()
+
+	ts := httptest.NewServer(writeHTML(`
+<body>
+	<p class="content">outer content</p>
+	<div id="section"><p class="content">inner content</p></div>
+</body>
+	`))
+	defer ts.Close()
+
+	var nodes []*cdp.Node
+	println("a")
+	if err := chromedp.Run(ctx,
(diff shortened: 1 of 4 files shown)


mvdan:
The CI failures seem unrelated; I'll have to investigate separately.

PR #644, opened 2020-06-16
introduce RunResponse to observe page navigations

### LINKED ISSUE #641: expose API to wait for a load event
Right now, the Navigate API implicitly waits for the page to load after issuing the page.Navigate command:

```
_, _, _, err := page.Navigate(urlstr).Do(ctx)
if err != nil {
        return err
}
return waitLoaded(ctx) 
```

However, we don't expose the functionality of `waitLoaded` in any way. It can be useful for many other scenarios which the user knows will trigger a page navigation, such as:

* clicking on a button
* running some javascript
* calling page.Navigate directly with custom parameters, such as https://github.com/chromedp/chromedp/issues/624
* expecting a page to trigger a redirect/navigation after some time

The simplest option would be to expose `waitLoaded` as-is. It's simple to understand and use - the moment it's called, it sets up a target listener and blocks until a load event is received. However, it's inherently racy; if the load event already passed by the time we set up the listener, we will probably block forever.

A better way would be to expose a `func(Action) Action` which wraps an action, doing the following in order when the returned action is executed:

* set up a listener
* run the underlying action
* wait until the load event is received

We could call this `chromedp.WaitLoad`, and it could be used like `chromedp.WaitLoad(chromedp.Click(xxx))`, or `chromedp.WaitLoad(page.Navigate(...).WithOption(...))`.

mvdan:
Until now, the API only provided a nice way to wait for navigations to
finish if they were directly triggered via Navigate. Doing the same for
a Click on a link, for example, required setting up ListenTarget and
manually waiting for the load event to show up.

Another common need was to check if a navigation ended up in error
status codes such as 404 or 503. Because of how a browser works, those
are still successful navigations, since the site might have a proper 404
landing page.

Fix both of those problems by adding a version of Run which waits for a
page navigation to finish, and returns the HTTP response information for
its document.

For more details, see the RunResponse docs, as well as
ExampleRunResponse.

Fixes #641.

### DIFF
--- a/chromedp.go
+++ b/chromedp.go
@@ -22,6 +22,7 @@ import (
 	"github.com/chromedp/cdproto/dom"
 	"github.com/chromedp/cdproto/inspector"
 	"github.com/chromedp/cdproto/log"
+	"github.com/chromedp/cdproto/network"
 	"github.com/chromedp/cdproto/page"
 	"github.com/chromedp/cdproto/runtime"
 	"github.com/chromedp/cdproto/target"
@@ -352,6 +353,7 @@ func (c *Context) attachTarget(ctx context.Context, targetID target.ID) error {
 	// Enable available domains and discover targets.
 	actions := []Action{
 		log.Enable(),
+		network.Enable(),
 	}
 	// These actions are not available on a worker target.
 	if !c.Target.isWorker {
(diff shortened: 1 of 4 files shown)


chromedp.go
```diff
@@ -409,6 +411,71 @@ func WithBrowserOption(opts ...BrowserOption) ContextOption {
...
+		if err := Run(ctx, actions...); err != nil {
+			return err
+		}
+
+		// Third, block until we have finished loading. If we wanted a
+		// response, check that it isn't nil.
+		select {
+		case <-ch:
+			if resp != nil && *resp == nil {
+				return fmt.Errorf("page loaded without a response")
```
rnd, line 470: nit: without a document type response since we only watch for `ResourceTypeDocument`

chromedp.go
```diff
@@ -409,6 +411,71 @@ func WithBrowserOption(opts ...BrowserOption) ContextOption {
 	}
 }
 
+// RunResponse is an alternative to Run which can be used with a list of actions
+// that trigger a page navigation, such as clicking on a link or button.
+//
+// RunResponse will run the actions and block until a page loads, returning the
+// HTTP response information for its HTML document. This can be useful to wait
```
rnd, line 418: not sure if the `ResourceTypeDocument` is only valid for HTML, how about any other document markup?

chromedp.go
mvdan: Good point :)

chromedp.go
mvdan: Hmm. I tried to think about this, but since we want to wait for the DOM document of the main frame, can that really be anything other than a Document?

I could add a test for loading a txt file, though, to see what happens there.

PR #663, opened 2020-07-15
Added ability to authentication proxies

webjohny:
Resolved a problem with authentication proxies. 
I used websocket connection without ready-made functions because there has delay. For this to work, you need to send a request without delay and queues.
Maybe it's seeming like hard code but it works. If you have any ideas for better realization please let me know)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,7 @@ cdp-*.txt
 /chromedp.test
 /chromedp.test.exe
 
+.idea
+
 /*.png
 /*.pdf
--- a/chromedp.go
+++ b/chromedp.go
@@ -11,15 +11,18 @@ package chromedp
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/chromedp/cdproto"
 	"github.com/chromedp/cdproto/browser"
 	"github.com/chromedp/cdproto/cdp"
 	"github.com/chromedp/cdproto/css"
 	"github.com/chromedp/cdproto/dom"
+	"github.com/chromedp/cdproto/fetch"
 	"github.com/chromedp/cdproto/inspector"
 	"github.com/chromedp/cdproto/log"
 	"github.com/chromedp/cdproto/network"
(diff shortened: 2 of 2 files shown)


mvdan:
Please add a test to show how this works; ideally, the test would spin up a local http proxy with authentication, so that we would test the real feature without having to use a proxy somewhere on the internet.

.gitignore
```diff
@@ -8,5 +8,7 @@ cdp-*.txt
 /chromedp.test
 /chromedp.test.exe
 
+.idea
```
mvdan, line 11: remove this; this should go in your global gitignore, not in all of your projects.

chromedp.go
```diff
@@ -385,6 +388,60 @@ func (c *Context) attachTarget(ctx context.Context, targetID target.ID) error {
...
+func Authentication(login string, password string) Tasks {
+	return Tasks{
+		fetch.Enable().WithPatterns([]*fetch.RequestPattern{{"*", "", ""}}).WithHandleAuthRequests(true),
+		ActionFunc(func(ctx context.Context) error {
+			if login != "" && password != "" {
+				c := FromContext(ctx)
+				if c == nil || c.Allocator == nil || c.cancel == nil {
+					return ErrInvalidContext
+				}
+				id := 1000
```
obalunenko, line 401: This ID looks like a magic number. Maybe it will be useful to have some comment that describes this variable, or even move this to constants?

chromedp.go
```diff
@@ -385,6 +388,60 @@ func (c *Context) attachTarget(ctx context.Context, targetID target.ID) error {
 	return nil
 }
 
+// Authentication proxy
+func Authentication(login string, password string) Tasks {
+	return Tasks{
+		fetch.Enable().WithPatterns([]*fetch.RequestPattern{{"*", "", ""}}).WithHandleAuthRequests(true),
+		ActionFunc(func(ctx context.Context) error {
+			if login != "" && password != "" {
```
obalunenko, line 396: Code could be simplified by decreasing nested constructions.
`if login == "" || password == ""{
return nil
}`

chromedp.go
```diff
@@ -385,6 +388,60 @@ func (c *Context) attachTarget(ctx context.Context, targetID target.ID) error {
...
+						})
+						cmd := &cdproto.Message{
+							ID:        int64(id + 1),
+							SessionID: c.Target.SessionID,
+							Method:    cdproto.MethodType("Fetch.continueWithAuth"),
+							Params:    buf,
+						}
+						err := c.Browser.conn.Write(ctx, cmd)
+						if err != nil {
+							panic(err)
```
obalunenko, line 422: Maybe it will be better to return error instead of panic?

chromedp.go
```diff
@@ -385,6 +388,60 @@ func (c *Context) attachTarget(ctx context.Context, targetID target.ID) error {
...
+						buf, _ := json.Marshal(map[string]string{"requestId":ev.RequestID.String()})
+						cmd := &cdproto.Message{
+							ID:        int64(id + 1),
+							SessionID: c.Target.SessionID,
+							Method:    cdproto.MethodType("Fetch.continueRequest"),
+							Params:    buf,
+						}
+						err := c.Browser.conn.Write(ctx, cmd)
+						if err != nil {
+							panic(err)
```
obalunenko, line 435: The same as above about panic usage

chromedp.go
```diff
@@ -385,6 +388,60 @@ func (c *Context) attachTarget(ctx context.Context, targetID target.ID) error {
...
+								"password": password,
+							},
+						})
+						cmd := &cdproto.Message{
+							ID:        int64(id + 1),
+							SessionID: c.Target.SessionID,
+							Method:    cdproto.MethodType("Fetch.continueWithAuth"),
+							Params:    buf,
+						}
+						err := c.Browser.conn.Write(ctx, cmd)
```
ZekeLu, line 420: It should not write raw messages here. `fetch.ContinueWithAuth` should be used instead.

ZekeLu:
The reviewing issues are not addressed. And I think it's not a good idea to implement it in `chromedp`. I have created a PR (https://github.com/chromedp/examples/pull/30) to show how to authenticate a proxy server. I will close this PR once that PR is merged.

@webjohny Please let me know if you think differently.

PR #666, opened 2020-07-19
sets c.allocated to nil on cancels

Oiyoo:
In Cancel, if ctx.Done's chan is read first then it's not a problem, but
if we read the allocated chan then any follow up cancel func calls will
block on their allocated chan reads.

Fixes by setting c.allocated to nil if the chan was read.

### DIFF
--- a/chromedp_test.go
+++ b/chromedp_test.go
@@ -1055,3 +1055,15 @@ func TestRunResponse_noResponse(t *testing.T) {
 		}
 	}
 }
+
+func TestCancelBlock(t *testing.T) {
+	ctx, cancel := NewContext(context.Background())
+	cancel()
+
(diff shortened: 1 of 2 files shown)


chromedp_test.go
```diff
@@ -1055,3 +1055,18 @@ func TestRunResponse_noResponse(t *testing.T) {
 		}
 	}
 }
+
+func TestCancelBlock(t *testing.T) {
+	alloCtx, cancel := NewExecAllocator(context.Background(), DefaultExecAllocatorOptions[:]...)
+	defer cancel()
```
mvdan, line 1061: do you need this? you could just do NewContext(context.Background()) I think

Oiyoo: done, thanks!

PR #667, opened 2020-07-22
support wait and after funcs with iframes

### LINKED ISSUE #72: consider APIs to help deal with iframes
Hi, 
Is there any way to set value in iframe and submit?

mvdan:
This makes use of Chrome's runtime execution context ID. In other words,
what javascript context we run our Evaluate commands in.

What we were doing before was locating a node's javascript object by
using its full XPath via the $x function. This worked fine for the most
part, but it did not work with iframes. Javascript isn't meant to have
access to the content of nested frames like that at all times, as it
would be a security concern.

Instead, in Chrome's console tab one can see a nifty "context" dropdown,
which defaults to "top", and allows you to run javascript in the context
of other frames in the page such as iframes.

We make use of that feature here, via the WithContextID option in the
Evaluate command. Each Frame has an ExecutionContextID, so we keep track
of those via runtime events and a map.

Then, when we use element objects via javascript, we make sure to use
the right execution context ID, depending on what frame we want to work
with. By default, that will be the top-level frame, as before; but if
the user specified FromNode, then that node's frame will be used
instead.

TestQueryIframe contains a sample of how this can work, and here is a
short sample:

        var iframes []*cdp.Node
        if err := chromedp.Run(ctx, chromedp.Nodes(`iframe`, &iframes, chromedp.ByQuery)); err != nil {
		handle(err)
	}

        var text string
        if err := chromedp.Run(ctx,
                chromedp.Text("#some_id", &text, chromedp.ByQuery, chromedp.FromNode(iframes[0])),
        ); err != nil {
                handle(err)
        }
	fmt.Println(text)

Fixes #72.

### DIFF
--- a/target.go
+++ b/target.go
@@ -2,6 +2,7 @@ package chromedp
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"sync"
 	"sync/atomic"
@@ -12,6 +13,7 @@ import (
 	"github.com/chromedp/cdproto/cdp"
 	"github.com/chromedp/cdproto/dom"
 	"github.com/chromedp/cdproto/page"
+	"github.com/chromedp/cdproto/runtime"
 	"github.com/chromedp/cdproto/target"
 )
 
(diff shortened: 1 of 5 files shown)


target.go
```diff
@@ -165,6 +174,44 @@ func (t *Target) Execute(ctx context.Context, method string, params easyjson.Mar
...
+		t.frameMu.Unlock()
+	case *runtime.EventExecutionContextsCleared:
+		t.frameMu.Lock()
+		for frameID := range t.execContexts {
+			delete(t.execContexts, frameID)
+		}
+		t.frameMu.Unlock()
+	}
+	_ = fmt.Print
+	// fmt.Println(t.execContexts)
```
Oiyoo, line 212: :wave:

mvdan: :facepalm:

PR #720, opened 2020-12-03
Do not set pdeathsig to the process on AWS Lambda

### LINKED ISSUE #562: Make allocateCmdOptions optional
I'm trying to run chromedp with Chromium (79.0.3945.0) in AWS Lambda but when `Run` is called for the first time and chromedp tries to spawn a new browser process, it fails with
```
fork/exec chromium: operation not permitted
```

This is caused by the call to `allocateCmdOptions` in `ExecAllocator.Allocate`, where it tries to set `pdeathsig` on the child Chrome process, which is (probably) not allowed in Lambda environment.

**Would it be possible to make the `allocateCmdOptions` call optional?** For FaaS environments like Lambda it serves no purpose anyway since if the main (chromedp) process dies, the whole environment (including the Chrome browser process) is destroyed and restarted.

dcu:
Since it's not supported by the AWS Lambda runtime

Fixes #562

### DIFF
--- a/allocate_linux.go
+++ b/allocate_linux.go
@@ -3,11 +3,18 @@
 package chromedp
 
 import (
+	"os"
 	"os/exec"
 	"syscall"
 )
 
 func allocateCmdOptions(cmd *exec.Cmd) {
+	_, isLambda := os.LookupEnv("LAMBDA_TASK_ROOT")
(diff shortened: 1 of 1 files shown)


ZekeLu:
Closing as being merged.

PR #730, opened 2020-12-22
Fix Browser goroutine/memory leak race

### LINKED ISSUE #552: potential heap memory leak when cancelling a browser context
```sh
github.com/chromedp/chromedp v0.4.2
Chromium 79.0.3945.117 Arch Linux
go version go1.13.5 linux/amd64
```

I am running `chromedp.Run` in endless loop. 
chromedp not flushing heap after context cancelled.

I copied [click](https://github.com/chromedp/examples/blob/master/click/main.go) example  and runned it in for loop. When I look at pprof, heap allocation increasing on every `chromedp.Run`

You can test at [http://localhost:8080/debug/pprof/](http://localhost:8080/debug/pprof/) after run the app.

```go
package main

import (
	"context"
	"log"
	"time"

	"github.com/gin-contrib/pprof"
	"github.com/gin-gonic/gin"
)

func main() {
	router := gin.Default()
	go func() {
		for {
			click()
		}
	}()
	pprof.Register(router)
	router.Run(":8080")

}

func click() {
	// create chrome instance
	ctx, cancel := chromedp.NewContext(
		context.Background(),
		chromedp.WithLogf(log.Printf),
	)
	defer cancel()

	// create a timeout
	ctx, cancel = context.WithTimeout(ctx, 15*time.Second)
	defer cancel()

	// navigate to a page, wait for an element, click
	var example string
	err := chromedp.Run(ctx,
		chromedp.Navigate(`https://golang.org/pkg/time/`),
		// wait for footer element is visible (ie, page is loaded)
		chromedp.WaitVisible(`body > footer`),
		// find and click "Expand All" link
		chromedp.Click(`#pkg-examples > div`, chromedp.NodeVisible),
		// retrieve the value of the textarea
		chromedp.Value(`#example_After .play .input textarea`, &example),
	)
	if err != nil {
		log.Fatal(err)
	}
	log.Printf("Go's time.After example:\n%s", example)
}


```

mpx:
Previously, the websocket read could race with context cancellation. The
(*Browser).run goroutine would return without signalling the allocator
cleanup goroutine to exit, which leaked a goroutine and prevented
garbage collecting the Browser memory.

Fixes #552

### DIFF
--- a/browser.go
+++ b/browser.go
@@ -225,13 +225,12 @@ func (b *Browser) run(ctx context.Context) {
 	// connection. The separate goroutine is needed since a websocket read
 	// is blocking, so it cannot be used in a select statement.
 	go func() {
+		// Signal to run and exit the browser cleanup goroutine.
+		defer close(b.LostConnection)
+
 		for {
 			msg := new(cdproto.Message)
 			if err := b.conn.Read(ctx, msg); err != nil {
-				// If the websocket failed, most likely Chrome was closed or
-				// crashed. Signal that so the entire browser handler can be
-				// stopped.
-				close(b.LostConnection)
 				return
 			}
 


mpx:
This test timeout appears to be unrelated?

PR #734, opened 2021-01-05
bug from chromedp/cdproto/target/target.go Do function 

TaroballzChen:
I modified the NewContext function from chromedp.go and it can work, if any problem, please let me know


kenshaw:
While I appreciate the work, this is literally a reversal of the most recent commit to `chromedp`. Unfortunately, since `cdproto` is completely generated, and Go is a statically compiled language, there are small snafus like this that will happen. It's not possible to detect what version of `cdproto` is being used and to use that instead. I had previously looked into tagging `cdproto` with the same revision versions as `chromium` itself, but this isn't really a viable option. Please update `chromedp` and `cdproto` and this issue will disappear for you. Thanks!

PR #737, opened 2021-01-13
Expose os.Process from the Browser

msoap:
Hi,
I've see #566, but we need a pid to control memory for all Chrome processes. We use this to export prometheus metrics with the consumed memory of sessions, for displaying on the Grafana dashboard.

Usage:
```
chromedp.FromContext(ctx).Browser.Process()
```

### DIFF
--- a/browser.go
+++ b/browser.go
@@ -119,6 +119,18 @@ func NewBrowser(ctx context.Context, urlstr string, opts ...BrowserOption) (*Bro
 	return b, nil
 }
 
+// Process returns the process object of the browser.
+// It could be nil when the browser is allocated with RemoteAllocator.
+// It can be used to control the memory consumed by the browser process.
+//
+// Example:
+//     if process := chromedp.FromContext(ctx).Browser.Process(); process != nil {
+//         fmt.Printf("Browser PID: %v", process.Pid)
+//     }
+func (b *Browser) Process() *os.Process {
+	return b.process
+}
+
 func (b *Browser) newExecutorForTarget(ctx context.Context, targetID target.ID, sessionID target.SessionID) (*Target, error) {
 	if targetID == "" {
 		return nil, errors.New("empty target ID")
@@ -308,15 +320,6 @@ func (b *Browser) run(ctx context.Context) {
 			return // to avoid "write: broken pipe" errors
 		}
 	}
-}
-
-// Process returns the Chrome process object
-// Example:
-//     if process := chromedp.FromContext(ctx).Browser.Process(); process != nil {
-//         fmt.Printf("Chrome PID: %v", process.Pid)
-//     }
-func (b *Browser) Process() *os.Process {
-	return b.process
 }
 
 // BrowserOption is a browser option.
--- a/browser.go
+++ b/browser.go
@@ -310,7 +310,7 @@ func (b *Browser) run(ctx context.Context) {
 	}
 }
 
-// Pid returning browser PID
+// Pid returns the browser process ID
 func (b *Browser) Pid() int {
 	return b.process.Pid
 }
--- a/browser.go
+++ b/browser.go
@@ -120,6 +120,7 @@ func NewBrowser(ctx context.Context, urlstr string, opts ...BrowserOption) (*Bro
 }
 
 // Process returns the process object of the browser.
+//
 // It could be nil when the browser is allocated with RemoteAllocator.
 // It can be used to control the memory consumed by the browser process.
 //
--- a/browser.go
+++ b/browser.go
@@ -310,9 +310,13 @@ func (b *Browser) run(ctx context.Context) {
 	}
 }
 
-// PID returns the browser process ID
-func (b *Browser) PID() int {
-	return b.process.Pid
+// Process returns the Chrome process object
+// Example:
+//     if process := chromedp.FromContext(ctx).Browser.Process(); process != nil {
+//         fmt.Printf("Chrome PID: %v", process.Pid)
+//     }
+func (b *Browser) Process() *os.Process {
+	return b.process
 }
 
 // BrowserOption is a browser option.
--- a/browser.go
+++ b/browser.go
@@ -122,7 +122,8 @@ func NewBrowser(ctx context.Context, urlstr string, opts ...BrowserOption) (*Bro
 // Process returns the process object of the browser.
 //
 // It could be nil when the browser is allocated with RemoteAllocator.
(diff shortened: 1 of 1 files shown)


browser.go
```diff
@@ -310,6 +310,11 @@ func (b *Browser) run(ctx context.Context) {
 	}
 }
 
+// Pid returning browser PID
```
dmitry-vovk, line 313:
```suggestion
// Pid returns the browser process ID
```

dmitry-vovk (approved):

browser.go
```diff
@@ -310,6 +310,11 @@ func (b *Browser) run(ctx context.Context) {
 	}
 }
 
+// Pid returns the browser process ID
+func (b *Browser) Pid() int {
```
dmitry-vovk, line 314: According to [Go Code Review requirement](https://github.com/golang/go/wiki/CodeReviewComments#initialisms):
```suggestion
func (b *Browser) PID() int {
```

dmitry-vovk (approved):

msoap:
@mvdan please take a look

msoap:
@kenshaw maybe you look?

browser.go
ZekeLu: I'm sorry, but I think it's better to make it consistant with the standard package, which is `Pid`. Or even better, call it `Getpid` (because this is a method instead of a field)? What do you think? @dmytro-vovk

browser.go
```diff
@@ -310,6 +310,11 @@ func (b *Browser) run(ctx context.Context) {
 	}
 }
 
+// PID returns the browser process ID
+func (b *Browser) PID() int {
+	return b.process.Pid
```
ZekeLu, line 315: Please note that the process could be `nil` when the browser instance is created by `chromedp.RemoteAllocator`. As commented by mvdan in #566, maybe it's better to export the process field. And can you provide the example to show how it will be used (we can add it to the doc comment)?

ZekeLu (changes requested):

browser.go
msoap: Agree, it makes sense, changed to expose Process, also added example to the doc comment

browser.go
```diff
@@ -310,6 +310,15 @@ func (b *Browser) run(ctx context.Context) {
 	}
 }
 
+// Process returns the Chrome process object
+// Example:
+//     if process := chromedp.FromContext(ctx).Browser.Process(); process != nil {
+//         fmt.Printf("Chrome PID: %v", process.Pid)
+//     }
```
ZekeLu, line 317: This example is good.
You mentioned that "need a pid to control memory for all Chrome processes", it would be better if you can elaborate on that.

browser.go
```diff
@@ -310,6 +310,15 @@ func (b *Browser) run(ctx context.Context) {
...
+func (b *Browser) Process() *os.Process {
```
ZekeLu, line 318: It seems that it's more natural to place this func after the func `NewBrowser`.

ZekeLu (changes requested):
I have added some more comments. Please take a look.
Sorry to quibble but let's try out best to make it better.

browser.go
```diff
@@ -310,6 +310,15 @@ func (b *Browser) run(ctx context.Context) {
 	}
 }
 
+// Process returns the Chrome process object
```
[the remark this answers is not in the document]
msoap: Done.

browser.go
```diff
@@ -310,6 +310,15 @@ func (b *Browser) run(ctx context.Context) {
...
+// Example:
+//     if process := chromedp.FromContext(ctx).Browser.Process(); process != nil {
+//         fmt.Printf("Chrome PID: %v", process.Pid)
```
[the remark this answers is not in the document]
msoap: Done.

browser.go
msoap: Done.

browser.go
msoap: Done.

browser.go
```diff
@@ -119,6 +119,18 @@ func NewBrowser(ctx context.Context, urlstr string, opts ...BrowserOption) (*Bro
 	return b, nil
 }
 
+// Process returns the process object of the browser.
+// It could be nil when the browser is allocated with RemoteAllocator.
+// It can be used to control the memory consumed by the browser process.
```
ZekeLu, line 124: I meant an example to show how to control the memory consumed by the browser process. It's not necessary to be put into the source code, it's okay to describe it in the pull request message.

Please note that #566 is abandoned because there is not a concrete purpose. We need an example to show that why we need to export this object.

ZekeLu (changes requested):

browser.go
```diff
@@ -119,6 +119,18 @@ func NewBrowser(ctx context.Context, urlstr string, opts ...BrowserOption) (*Bro
 	return b, nil
 }
 
+// Process returns the process object of the browser.
+// It could be nil when the browser is allocated with RemoteAllocator.
```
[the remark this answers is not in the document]
msoap: Done.

browser.go
msoap: Just for monitoring. We use this to export prometheus metrics with the consumed memory of sessions, for displaying on the Grafana dashboard.

ZekeLu: It makes sense now. Thank you very much! :+1:

ZekeLu: Let's change the line to:

```go
// It could be useful for a monitoring system to collect process metrics of the browser.
// (see https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#NewProcessCollector for an example).
```

What do you think?

ZekeLu (changes requested):

browser.go
ZekeLu: Let's change this line to:

```go
// It could be useful for a monitoring system to collect process metrics of the browser process.
// (see https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#NewProcessCollector for an example)
```

What do you think?

msoap: Changed

ZekeLu (approved):
Hey, @msoap It looks good to me now. Thank you very much!

Hi, @kenshaw Do you have time to take a look?

kenshaw:
@ZekeLu I'm fine with this. Absolutely no problem exposing lower internals in general.

ZekeLu:
Thank you all! I'm going to merge this request.

PR #743, opened 2021-01-18
Wrap errors with go 1.13 %w verb

sashayakovtseva:
Hello,

I am using browserless and connecting to it with chromedp. When browserless is under a high load ws connections are closed (see [this](https://docs.browserless.io/docs/docker.html#max-queue-length) option) and dial returns `io.EOF` after the first read. To make it more convenient I suggest to wrap dial errors to enable `errors.Is` behavior instead of direct strings comparison here.


ZekeLu:
Hi @sashayakovtseva , would you like to find all the `fmt.Errorf` and fix them together? For example, #793 should be fixed too.

sashayakovtseva:
Hey @ZekeLu,

I've just pushed changes for wrapping errors with the `%w` verb. However, in #793 `page load error` wraps yet another string. 

I'd rather leave that kind of refactoring to maintainers since that is a pretty large piece I am not willing to change now, sorry.

ZekeLu:
Hello, @sashayakovtseva . My bad, I thought they are the same.

Just double checked, you've handled all the `%w` case. Thank you!

Issue #680, opened 2020-08-07
There's currently no way to wait for a condition except with WaitFunc and a selector


opennota:
What I want is to evaluate some JS repeatedly until it returns the expected result. There seems to be no other way to do that but with a clumsy code like this, using `WaitFunc` and a random selector:

``` Go
chromedp.Query("random-selector",
	func(s *chromedp.Selector) {
		chromedp.WaitFunc(func(ctx context.Context, cur *cdp.Frame, ids ...cdp.NodeID) ([]*cdp.Node, error) {
			if len(ids) == 0 {
				return nil, nil
			}
			var result string
			if err := chromedp.Evaluate(someJavascript, &result).Do(ctx); err != nil {
				return nil, err
			} else if result != whatWeAreWaitingFor {
				return nil, nil
			}
			return []*cdp.Node{}, nil
		})(s)
	},
	chromedp.AtLeast(0),
)
```

It would be great if one could wait for a JS condition using a simpler action.

opennota:
FWIW, puppeteer has `waitForFunction`:

https://devdocs.io/puppeteer/index#pagewaitforfunctionpagefunction-options-args

yoyo00xx:
There should be something like the puppeteer API

ZekeLu:
@opennota It seems that you just want the **wait** part of `chromedp.Query`. It's implemented like this:

https://github.com/chromedp/chromedp/blob/e2970556e3d05f3259c464faeed1ec0e862f0560/query.go#L161-L166

Borrow this implementation, your code can be changed to (I have extended it to be runnable, and it's based on `chromedp v0.6.5`):

```go
package main

import (
	"context"
	"fmt"
	"log"
	"net/http"
	"net/http/httptest"
	"time"

	"github.com/chromedp/chromedp"
)

func main() {
	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		fmt.Fprint(w, `
<html>
<script>
let v = 0;
setTimeout(()=>{v = 100}, 1000);
</script>
<body>
hello world
</body>
</html>
`)
	}))
	defer ts.Close()

	ctx, cancel := chromedp.NewContext(context.Background(), chromedp.WithDebugf(log.Printf))
	defer cancel()
	ctx, cancel = context.WithTimeout(ctx, 5*time.Second)
	defer cancel()

	if err := chromedp.Run(ctx,
		chromedp.Navigate(ts.URL),
[...]

PR #766, opened 2021-03-08
add poll action to imitate puppeteer's page.waitForFunction

ZekeLu:
To fix issue #680.

The new action is called `Poll` to avoid messing up with the `Wait*` query actions.

Here is the documentation for [Puppeteer's page.waitForFunction](https://github.com/puppeteer/puppeteer/blob/v8.0.0/docs/api.md#pagewaitforfunctionpagefunction-options-args).

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -45,4 +45,7 @@ const (
 
 	// ErrInvalidContext is the invalid context error.
 	ErrInvalidContext Error = "invalid context"
+
+	// ErrPollingTimeout is the error that the timeout reached before the pageFunction returns a truthy value.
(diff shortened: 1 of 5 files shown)


ysmood:
One trade-off, the retry can't survive page reload. Such as the page keeps redirecting to different URLs after we click a button, we want to wait until another special button shows up on the page.

The actual performance gain of doing the retry inside the browser is usually trivial.

ZekeLu:
Please ignore this pull request for now, because the tests will fail sometimes. I will try to figure out what's wrong.

kenshaw:
Merged.

PR #776, opened 2021-03-21
make tests run in Github Actions pass

ZekeLu:
1. it can not write to the `/tmp` directory. We should use `RUNNER_TEMP` as the temp directory.
2. `TestExecAllocatorMissingWebsocketAddr` got different error in Github Actions:
    local:
    ```
    chrome failed to start:
    [0321/235654.258252:ERROR:headless_shell.cc(720)] Invalid devtools server address
    ```
    github actions:
    ```
    chrome failed to start:
    /bin/bash: /etc/profile.d/env_vars.sh: Permission denied
    mkdir: cannot create directory ‘/run/user/1001’: Permission denied
    [0321/081807.491906:ERROR:headless_shell.cc(720)] Invalid devtools server address
    ```
3. `TestKeyEvent` and `TestKeyEventNode` will crash `headless-shell`, so skip them until [the issue](https://bugs.chromium.org/p/chromium/issues/detail?id=1170634&q=IsSupportedClipboardBuffer&can=2) has been fixed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,4 +32,4 @@ jobs:
     - name: Build
       run: go test -c -vet=off
     - name: Test
-      run: docker run --rm --volume=$PWD:/chromedp --entrypoint=/chromedp/chromedp.test --workdir=/chromedp --env=PATH=/headless-shell chromedp/headless-shell:latest -test.v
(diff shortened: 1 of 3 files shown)


kenshaw:
Merged.

Issue #722, opened 2020-12-04
fetch.FulfillRequest when body too big，the chromium crashed.


GuangxinZhang:
#### What versions are you running?

<pre>
$ go list -m github.com/chromedp/chromedp
github.com/chromedp/chromedp v0.5.3
$ google-chrome --version
Chromium 78.0.3870.0 stable
$ go version
go version go1.14.1 linux/amd64
</pre>

#### What did you do? Include clear steps.
I use fetch.FulfillRequest to modify the http response:
<pre>
err := fetch.FulfillRequest(ev.RequestID, int64(status)).
    WithBody(base64.StdEncoding.EncodeToString([]byte(respBody))).
    WithResponseHeaders(headers).
    WithResponsePhrase("OK").
    Do(ctx)
</pre>
but when the respBody is too big (961kb), the chromium crashed!

2020/12/04 15:23:38 ERROR: write tcp 127.0.0.1:53546->127.0.0.1:37910: write: broken pipe

ysmood:
From Rod's [readme](github.com/go-rod/rod):

> Chromedp uses third part WebSocket lib which has [1MB overhead](https://github.com/chromedp/chromedp/blob/b56cd66f9cebd6a1fa1283847bbf507409d48225/conn.go#L43-L54) for each cdp client, if you want to control thousands of remote browsers it can become a problem. Because of this limitation, if you evaluate a js script larger than 1MB, Chromedp will crash.

[Rod](github.com/go-rod/rod) doesn't have this limitation: https://gist.github.com/ysmood/0d5b2c878ecbdb598776af7d3d305b79

Here's how Rod solves it, might help chromedp to improve the way it uses websocket: https://github.com/go-rod/rod/blob/master/lib/cdp/websocket.go

GuangxinZhang:
> From Rod's [readme](github.com/go-rod/rod):
> 
> > Chromedp uses third part WebSocket lib which has [1MB overhead](https://github.com/chromedp/chromedp/blob/b56cd66f9cebd6a1fa1283847bbf507409d48225/conn.go#L43-L54) for each cdp client, if you want to control thousands of remote browsers it can become a problem. Because of this limitation, if you evaluate a js script larger than 1MB, Chromedp will crash.
> 
> Rod doesn't have this limitation: https://gist.github.com/ysmood/0d5b2c878ecbdb598776af7d3d305b79
> 
[...]

Issue #401, opened 2019-06-26
sending large messages to Chrome via websocket isn't well supported


mvdan:
This is a placeholder issue so that others can easily find it and chime in in the future. For context, see https://github.com/ChromeDevTools/devtools-protocol/issues/175.

To help chromedp users debug this issue, I'll make chromedp error with a link to this issue for now.

Like the issue linked above says, there isn't a good solution until Chrome supports fragmentation. For now, trying to send large messages will simply error. Making the write buffer much larger is arguably a worse situation to be in.

kenshaw:
This is also a reason to move towards using the files instead of sockets. We can also allow users to change the websocket dialer.

Issue #395, opened 2019-06-18
sendMessageToTarget with large payloads causes "context canceled"


tmm1:
#### What versions are you running?

<pre>
$ go list -m github.com/chromedp/chromedp
github.com/chromedp/chromedp v0.3.0
$ google-chrome --version
Google Chrome 75.0.3770.100
$ go version
go version go1.12.4 darwin/amd64
</pre>

#### What did you do? Include clear steps.

I'm trying to implement a simple cookie jar between chromedp runs.

At the end of a session I use `network.GetAllCookies` to save a list of `[]*network.Cookie`

At the beginning of a new session, I first convert my `[]*network.Cookie` to `[]*network.CookieParam`, then call `network.SetCookies`.

#### What did you expect to see?

I expect `network.SetCookies` to return no error.

#### What did you see instead?

For 8 cookies and 12 cookies, things work as expected.

But when I had 20 cookies, `network.SetCookies` is returning `context canceled`

mvdan:
Can you provide a way to reproduce the error? Without that, it's hard to tell what's actually going on.

Also see the FAQ in the readme; one common way to get unexpected `context canceled` errors is because Chrome unexpectedly died and the websocket connection was lost. So looking at what happened with your chrome process could help. We will probably give better errors in the future, but for now it shouldn't be too hard to manually debug.

tmm1:
> Also see the FAQ in the readme; one common way to get unexpected `context canceled` errors is because Chrome unexpectedly died and the websocket connection was lost.

Thanks for this pointer. It does seem like chrome is crashing, but I can't tell why.

I tried using `chromedp.WithDebugf(log.Printf)` but all I see is the last `Target.sendMessageToTarget` that triggers the crash. Is there some way to make chrome more verbose about what it dislikes about my message?

tmm1:
I tried using this to enable logs, but I still get no logs to stderr or in the user-data-dir on my mac:

```
opts = append(opts,
    chromedp.Flag("enable-logging", "stderr"),
    chromedp.Flag("log-level", "0"),
    chromedp.Flag("v", "1"),
)
```

However, I was able to confirm my hunch that this is related to the size of the message sent to chrome. Here is a trivial reproduction:

```
chromedp.ActionFunc(func(ctx context.Context) error {
	_, _, err := runtime.Evaluate(fmt.Sprintf("//%s", strings.Repeat("x", 4096))).Do(ctx)
	return err
}),
```

[...]

PR #782, opened 2021-03-31
auto grows the buffer to workaround the lack of fragmentation support

ZekeLu:
To workaround this issue: https://github.com/chromedp/chromedp/issues/401

And the following  related issues:

* https://github.com/chromedp/chromedp/issues/395
* https://github.com/chromedp/chromedp/issues/722

According to https://bugs.chromium.org/p/chromium/issues/detail?id=1069431, it looks like that we won't get fragmentation support in the near future. Meanwhile, the underlying package https://github.com/gobwas/ws adds auto growing feature recently, so that we can address the issue from the client side.

Please note that this PR is not ready for merged yet, because it uses the new feature in the RC version, and the new feature has some issues (see https://github.com/gobwas/ws/pull/134).


kenshaw:
Merged.

PR #778, opened 2021-03-25
Fix some of the issues in the Poll action

ZekeLu:
The Poll action failed the tests very often. I have found two issues in the implementation:

1. it does not wait for the `runtime.ExecutionContextID` to be available.
2. the tests for the `WithPollingMutation()` option is unreliable.

This PR fixes those issues.

@kenshaw Sorry to ping you directly, but can you take some time to review this changes? The buggy implementation has already been merged into the master branch, I hope that the issues can be addressed asap. Thank you very much!

PR #786, opened 2021-04-03
close timer appropriately

robinbraemer:


robinbraemer:
don't know why tests failed

ZekeLu:
The Poll action has some issues, and those issues have been addressed in https://github.com/chromedp/chromedp/pull/778. Waiting to be merged.

BTW, do you mind to explain why the change in your PR? It looks the same to me.

robinbraemer:
> BTW, do you mind to explain why the change in your PR? It looks the same to me.

Sure, so the Timer would only be stopped within the first `case` but not the second. We always want to stop the Timer if its not needed anymore. I deferred  `t.Stop()` which closes the Timer in any case which makes more sense.

PR #787, opened 2021-04-06
add the MouseMoveNode function

rn0l485:
Hi, 
I just add a function for moving the mouse to the target node.


input.go
```diff
@@ -23,6 +23,28 @@ func MouseEvent(typ input.MouseType, x, y float64, opts ...MouseOption) MouseAct
 	return p
 }
 
+
+
+
+
```
ZekeLu, line 29: remove extra blank lines

ZekeLu, line 29: remove extra blank lines

input.go
```diff
@@ -49,6 +71,38 @@ func MouseClickXY(x, y float64, opts ...MouseOption) MouseAction {
 	})
 }
 
+
+
+
+
```
ZekeLu, line 77: remove extra blank lines

input.go
```diff
@@ -49,6 +71,38 @@ func MouseClickXY(x, y float64, opts ...MouseOption) MouseAction {
...
+		x /= float64(c / 2)
+		y /= float64(c / 2)
+
+		return MouseMoveXY(x, y, opts...).Do(ctx)
+	})
+}
+
+
+
+
```
ZekeLu, line 105: remove extra blank lines above.

input.go
```diff
@@ -23,6 +23,28 @@ func MouseEvent(typ input.MouseType, x, y float64, opts ...MouseOption) MouseAct
...
+func MouseMoveXY(x, y float64, opts ...MouseOption) MouseAction {
+	return ActionFunc(func(ctx context.Context) error {
+		p := &input.DispatchMouseEventParams{
+			Type:       input.MouseMoved,
+			X:          x,
+			Y:          y,
+		}
```
ZekeLu, line 36: better to use the constructor func?

```go
p := input.DispatchMouseEvent(input.MouseMoved, x, y)
```

input.go
```diff
@@ -49,6 +71,38 @@ func MouseClickXY(x, y float64, opts ...MouseOption) MouseAction {
 	})
 }
 
+
+
+
+
+func MouseMoveNode(n *cdp.Node, opts ...MouseOption) MouseAction {
+	return ActionFunc(func(ctx context.Context) error {
```
ZekeLu, line 79: `MouseClickNode` do some check and scroll the node into view, should we do the same here? If the answer is yes, maybe we should  extract those code into a func.

ZekeLu (changes requested):
And can you please add unit tests  for the new funcs?

input.go
rn0l485: The MouseMoveNode function is the similar function to the MouseClickNode, but it doesn’t do the click movement, just put the mouse on the target node.

input.go
rn0l485: Yes, I just noticed this function here, it did the thing what I’m looking for for MouseMoveXY

ysmood:
@rn0l485 A better way to abstract it in Rod: https://github.com/go-rod/rod/blob/b69b18c93345bdb71e824a2fe579de5483a221e3/element.go#L71-L90

```go
page.MustElement("button").MustHover()
```

PR #809, opened 2021-05-02
Add allocator option: IgnoreCertErrors

shelld3v:


ZekeLu:
Thanks for the contribution!

There are lots of command line options for the Chrome browser ([a non-official and a little outdated list](https://peter.sh/experiments/chromium-command-line-switches/)). I'm not sure whether this one is special enough to get a name. And if we want to give it a name, I think `IgnoreCertErrors` looks more accurate.

ZekeLu:
It turns out that this option is useful (I used it several times when using a proxy to capture the network traffics), and the cited web page does not include this option, so I'm going to merge this PR.

shelld3v:
@ZekeLu: It turns out that this option is useful, so I'm going to merge this PR
Also @ZekeLu: *closed #809*

ZekeLu:
Sorry for making you confuse. I want to move this option after `ProxyServer` and modify the comment, so I can't merge this PR directly. This PR is closed after the cherrypick. You can find the contribution in the commit history. Thank you!

Issue #502, opened 2019-10-30
Unable to use chromedp in docker-compose


donfrigo:
#### What versions are you running?

Ubuntu 18.04
Go 1.13.3 linux/amd64

#### What did you do? Include clear steps.

I have added one of the simple example projects to my main google file and set up the docker the following way.

Dockerfile:

    RUN go get -u github.com/golang/dep/cmd/dep

    ADD ./main.go /go/src/app
    COPY ./Gopkg.toml /go/src/app

    WORKDIR /go/src/app

    RUN dep ensure
    RUN go build

    CMD ["./app"] 

Docker-compose:

    version: '3'
    services:
      chromedp:
        image: chromedp/headless-shell
        ports:
          - 9222:9222
      goservice:
        build: .
        links:
          - chromedp
        depends_on:
          - chromedp`

Main Go file:
```
package main

import (
	"context"
	"log"
	"time"

	"github.com/chromedp/chromedp"
)

func main() {
	// create chrome instance
	ctx, cancel := chromedp.NewContext(
		context.Background(),
		chromedp.WithLogf(log.Printf),
	)
	defer cancel()

	// create a timeout
	ctx, cancel = context.WithTimeout(ctx, 15*time.Second)
	defer cancel()

	// navigate to a page, wait for an element, click
	var example string
	err := chromedp.Run(ctx,
		chromedp.Navigate(`https://golang.org/pkg/time/`),
		// wait for footer element is visible (ie, page is loaded)
		chromedp.WaitVisible(`body > footer`),
		// find and click "Expand All" link
		chromedp.Click(`#pkg-examples > div`, chromedp.NodeVisible),
		// retrieve the value of the textarea
		chromedp.Value(`#example_After .play .input textarea`, &example),
	)
	if err != nil {
		log.Fatal(err)
	}
	log.Printf("Go's time.After example:\n%s", example)
}
```

#### What did you expect to see?

According to the FAQ section of the chromedp library it should be pretty straightforward to set it up. 

> The simplest way is to run the Go program that uses chromedp inside the chromedp/headless-shell image. That image contains headless-shell, a smaller headless build of Chrome, which chromedp is able to find out of the box.

So I expected to see the output of the parsed text upon running

`docker-compose up --build`

#### What did you see instead?

`goservice_1  | 2019/10/30 18:37:42 exec: "google-chrome": executable file not found in $PATH`

[...]

PR #817, opened 2021-05-11
make `NewRemoteAllocator` accept url without `devtools/browser/...`

### LINKED ISSUE #438: Connecting to docker chromedp/headless-shell instance via RemoteAllocator
I would like to know how to connect to a chromedp/headless-shell docker instance as a remote allocator. I have seen a few comments recommending to run the go binary inside of the chromedp/headless-shell container as a solution, but it has also been implied this is not the only way to connect. Is it possible to have a go process connect to headless-shell **through** a running docker instance of chromedp/headless-shell but with the go binary running on the host instead of inside the headless-shell container?

The chromedp/headless-shell documentation gives this run command as an example: `docker run -d -p 9222:9222 --rm --name headless-shell chromedp/headless-shell` which I can only assume is given for the purpose of connecting to the instance from outside, is this correct?

It seems to me this was the intention of the RemoteAllocator interface. But the problem I am having is that I don't know what `url string` to pass to `func NewRemoteAllocator(parent context.Context, url string)`. I have tried passing it `ws://127.0.0.1:9222` but it errors with `could not dial "ws://127.0.0.1:9222": unexpected HTTP response status: 404`. I know the url should probably look something like `ws://127.0.0.1:9222/devtools/browser/12939166-3db0-49cd-9aaa-56381ef91a7f` but the last portion is dynamic and I do not know how to obtain that url from the docker instance running headless-shell.

ZekeLu:
A full websocket debugger URL is something like this:

```
ws://127.0.0.1:9222/devtools/browser/120e6fb7-0094-47a2-8b4c-51741797f280
```

And the ID part is not static, which requires extra work to get the correct URL for `chromedp.NewRemoteAllocator`.

This change allows the client to just pass the base URL to `chromedp.NewRemoteAllocator`. It will send a request to `/json/version`  to get the full websocket debugger URL if the provided URL is not  a valid websocket debugger URL.

After the change, the URL with the following formats are accepted:
* `ws://127.0.0.1:9222/`
* `http://127.0.0.1:9222/`

Do not use URL in the following formats because they contain `/devtools/browser/` and will be considered as valid websocket debugger URLs:
* `ws://127.0.0.1:9222/devtools/browser/`
* `ws://127.0.0.1:9222/devtools/browser/`
* `http://127.0.0.1:9222/devtools/browser/120e6fb7-0094-47a2-8b4c-51741797f280`

Fixes #438.

### DIFF
--- a/allocate.go
+++ b/allocate.go
@@ -489,13 +489,9 @@ func CombinedOutput(w io.Writer) ExecAllocatorOption {
 // Because it contains "/devtools/browser/" and will be considered
 // as a valid websocket debugger URL.
 func NewRemoteAllocator(parent context.Context, url string) (context.Context, context.CancelFunc) {
-	url, err := detectURL(url)
-	if err != nil {
-		panic(fmt.Sprintf("failed to detect the websocket debugger url: %v", err))
-	}
 	ctx, cancel := context.WithCancel(parent)
 	c := &Context{Allocator: &RemoteAllocator{
-		wsURL: url,
+		wsURL: detectURL(url),
 	}}
 	ctx = context.WithValue(ctx, contextKey{}, c)
 	return ctx, cancel
--- a/allocate.go
+++ b/allocate.go
@@ -478,7 +478,21 @@ func CombinedOutput(w io.Writer) ExecAllocatorOption {
 // NewRemoteAllocator creates a new context set up with a RemoteAllocator,
 // suitable for use with NewContext. The url should point to the browser's
 // websocket address, such as "ws://127.0.0.1:$PORT/devtools/browser/...".
+// If the url does not contain "/devtools/browser/", it will try to detect
+// the correct one by sending a request to "http://$HOST:$PORT/json/version".
+//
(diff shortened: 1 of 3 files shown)


ZekeLu:
This PR also takes care of #502. Tested with the following configuration:

`docker-compose.yml`:

```yml
version: "3.9"
services:
  browser:
    image: chromedp/headless-shell:88.0.4324.182
  app:
    image: testapp:latest
    links:
      - browser
```

`main.go`:

```go
package main

import (
	"context"
	"log"

)

func main() {
	// pay attention to the url below
	ctx, cancel := chromedp.NewRemoteAllocator(context.Background(), "ws://browser:9222/")
	defer cancel()

	ctx, cancel = chromedp.NewContext(ctx, chromedp.WithDebugf(log.Printf))
	defer cancel()

	if err := chromedp.Run(ctx,
		chromedp.Navigate("https://www.google.com/"),
	); err != nil {
		log.Fatal(err)
	}
}
```

PR #962, opened 2021-11-30
add another version google chrome for macos

rsmnarts:
Previously, I found an error when using another version of google chrome for macOS.


ZekeLu:
Sorry but I don't think that it is necessary to list all the possible paths. There is an option to specify the path:

https://github.com/chromedp/chromedp/blob/master/allocate.go#L323-L326

PR #963, opened 2021-11-30
Add support for `profile-directory` command line option

zvodd:


allocate.go
```diff
@@ -424,6 +424,17 @@ func UserDataDir(dir string) ExecAllocatorOption {
...
 }
 
+// UserDataProfileDir is the command line option to set the profile
+// subdirectory under the UserData directory.
+//
+// Note: if unspecificed Chrome will choose/create the folder "Default".
+// This folder contains the profile specific data like cookies & cache.
+// If multiple profile exists they are typically named "Profile 1",
+// "Profile 2", etc.
+func UserDataProfile(profile string) ExecAllocatorOption {
```
ZekeLu, line 434: I think `ProfileDirectory` is better.

allocate.go
```diff
@@ -424,6 +424,17 @@ func UserDataDir(dir string) ExecAllocatorOption {
 	return Flag("user-data-dir", dir)
 }
 
+// UserDataProfileDir is the command line option to set the profile
```
ZekeLu, line 427: - `UserDataProfileDir` --> `ProfileDirectory`
- `the profile subdirectory under the UserData directory` --> `the profile directory under  the user data directory`

allocate.go
zvodd: Yeah I was a little bit concerned about the ambiguity there. Because the word "directory" might imply that you need to specify a full path.
I think that this is a actually a rare enough use case as well as being potentially confusing enough, to cancel this pull request.

If someone needs the feature they can always just use `chromedp.Flag`:
```
opts := []chromedp.ExecAllocatorOption{
	chromedp.UserDataDir("/home/user/.config/google-chrome"),
	chromedp.Flag("profile-directory", "Profile 2"),
}
ctx, cancel := chromedp.NewExecAllocator(context.Background(), opts...)
defer cancel()
ctx, cancel = chromedp.NewContext(ctx)
defer cancel()
```

I think maybe providing an example in examples might be a better alternative.

ZekeLu: > I think that this is a actually a rare enough use case as well as being potentially confusing enough, to cancel this pull request.

Yes, I can not think of a use case for this option.

PR #1032, opened 2022-03-11
Search for Flatpak binaries and ungoogled-chromium

dawidpotocki:


ZekeLu:
Thanks for the pull request! But in my opinion, those paths don't have to go into the lib. The users can specify a path that is not built-in easily:

```go
	opts := append(chromedp.DefaultExecAllocatorOptions[:],
		chromedp.ExecPath("path to the browser"),
	)
	allocCtx, cancel := chromedp.NewExecAllocator(context.Background(), opts...)
	defer cancel()
	ctx, cancel := chromedp.NewContext(allocCtx)
	defer cancel()
```

dawidpotocki:
Well, I saw that snap is included, so why not Flatpak?
Unlike snap it's not used only by Ubuntu.

ZekeLu:
FYI, chrome is still on flathub-beta and not graduated yet: https://discourse.flathub.org/t/google-chrome-on-flathub-beta/634

I will leave the decision to @kenshaw .

dawidpotocki:
Chrome just got promoted to Flathub stable.
https://www.flathub.org/apps/details/com.google.Chrome
https://www.flathub.org/apps/details/com.google.ChromeDev

Issue #1031, opened 2022-03-11
Go get error


HewieWang:
#### What versions are you running?

```console
$ go list -m github.com/chromedp/chromedp
p@v0.7.8
$ google-chrome --version
96.0.4664.93
$ go version
go version go1.17.6 windows/amd64
```

#### What did you do? Include clear steps.
```go
go get github.com/chromedp/chromedp
```

#### What did you see instead?

..\..\go\pkg\mod\github.com\chromedp\chromedp@v0.7.8\query.go:1139:8: assignment mismatch: 7 variables but css.GetMatchedStylesForNode(nodes[0].NodeID).Do returns 8 values

victor-u:
Me too

ZekeLu:
Thanks for reporting the issue. Please downgrade `github.com/chromedp/cdproto` to match `github.com/chromedp/chromedp@v0.7.8` for the time being:

```sh
go get github.com/chromedp/cdproto@v0.0.0-20220217222649-d8c14a5c6edf
```

ZekeLu:
Addressed by https://github.com/chromedp/chromedp/commit/b88710e33da89f65c9ed750381125aea3922254c

PR #1035, opened 2022-03-14
Fix query.go assignment mismatch

StevenGutzwiller:
Fix for https://github.com/chromedp/chromedp/issues/1031


ZekeLu:
Addressed by https://github.com/chromedp/chromedp/commit/b88710e33da89f65c9ed750381125aea3922254c

PR #1083, opened 2022-06-12
feat: add pointer support to AtLeast

johnrichardrinehart:
# Description

I had a problem, recently, where I needed to use the value of the number of elements discovered after a page had loaded in an `AtLeast` option. With the current (official) implementation of `AtLeast` I was unable to chain together `Action`s that used the number discovered on page load. I have used this PR's implementation with code like the below and it seems to work, but I haven't fixed the tests, yet. So, if the concept of the PR is approved then I'll work on fixing the tests in this branch, too.

```go
	var numberOfItemsString string
	numberOfItems := 0
	actions := []chromedp.Action{
		chromedp.Navigate(url),
		chromedp.Text(`//*[@id="app"]/div[2]/div[3]/div[1]/div[2]/div[1]/div[1]`, &numberOfItemsString),
		// scroll to bottom
		// https://github.com/chromedp/chromedp/issues/525#issuecomment-637058121
			time.Sleep(2 * time.Second)
			_, exp, err := runtime.Evaluate(`window.scrollTo(0,document.documentElement.scrollHeight);`).Do(ctx)
			if err != nil {
				return err
			}
			if exp != nil {
				return exp
			}
			return nil
		}),
		// write number of nodes to variable
			n, err := strconv.Atoi(numberOfItemsString)
			if err != nil {
				return fmt.Errorf("failed to convert items string %s to integer: %s", numberOfItemsString, err)
			}

			numberOfItems = n
			return nil
		}),
	}
	
	var ns []*cdp.Node
	xp := `//*[@id="asset-card"]/div[1]/img`
	actions = append(actions,
		chromedp.Nodes(xp, &ns, chromedp.BySearch, chromedp.AtLeast(&numberOfItems)),
	)
	
```

### DIFF
--- a/query.go
+++ b/query.go
@@ -28,7 +28,7 @@ type QueryAction Action
 type Selector struct {
 	sel      interface{}
 	fromNode *cdp.Node
-	exp      int
+	exp      *int
 	by       func(context.Context, *cdp.Node) ([]cdp.NodeID, error)
 	wait     func(context.Context, *cdp.Frame, runtime.ExecutionContextID, ...cdp.NodeID) ([]*cdp.Node, error)
 	after    func(context.Context, runtime.ExecutionContextID, ...*cdp.Node) error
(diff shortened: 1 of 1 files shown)


johnrichardrinehart:
@ZekeLu's suggestion in #1084 preempts the need for this solution. Closing in favor of preserving the API.

Issue #922, opened 2021-10-05
Cannot use pageNumber / totalPages with chromedp/docker-headless-shell


boindil:
#### What versions are you running?

Since the other repo does not allow issues to be created, I am trying it here.

#### What did you do? Include clear steps.

I am using a go service based on chromedp that is creating PDFs. I included pageNumber and totalPages in the footer.

When using `FROM chromedp/headless-shell:stable`, I'm getting the behaviour described below. When using `FROM debian:stable-slim` along with `RUN apt-get install -y chromium`, everything is working is expected.

Is this a bug in the original `headless-shell`, the modifications made in the `chromedp/headless-shell` or whats happening here?

#### What did you expect to see?

I expected the pagenumber and totalpages to be visible in the footer

#### What did you see instead?

Nothing

ZekeLu:
Please provide a minimal reproducer which creates PDFs with footer.

ZekeLu:
@boindil Thanks for reporting the issue. The cause is the lack of the `locals` folder in the `chromedp/headless-shell` image. A workaround before `chromedp/headless-shell` is fixed is to build a customized image yourself:

```Dockerfile
FROM debian:buster-slim AS chromium
RUN apt-get update -y && apt-get install -y chromium

FROM chromedp/headless-shell:stable AS final
COPY --from=chromium /usr/lib/chromium/locales /headless-shell/locales
```

Contents of the `locales` folder:

```shell
$ ls -lh locales/
total 252K
-rw-r--r-- 1 root root 252K May 15 20:39 en-US.pak
```

@kenshaw I think we should include the `locales` folder in the `chromedp/headless-shell` image. Does it exist in the `out` directory after the build? Or should we copy it from somewhere else?

kenshaw:
@boindil actually, yes I didn't see this. I've aware of the missing `locales` folder issue, as it's come up previously. I'll change the build script to incorporate the locales. Might be a bit before I get to this, as I am deep in other projects at the moment.

fadeer:
```Dockerfile
FROM debian:buster-slim AS chromium
RUN apt-get update -y && apt-get install -y chromium

FROM chromedp/headless-shell:stable AS final
COPY --from=chromium /usr/lib/chromium/locales /headless-shell/locales
```

Thank you for this example, locales is also required by Grafana UI. If not includes, console show exception:

```
exception "Uncaught" (1:2894997): RangeError: Incorrect locale information provided
    at new NumberFormat (<anonymous>)
[...]

PR #1086, opened 2022-06-13
add a test to make sure resource pack is loaded in `headless-shell`

ZekeLu:
For #922


ZekeLu:
The failure is expected. It uncovers an issue in the `chromedp/headless-shell` image.

The PR should be merged after chromedp/docker-headless-shell#19 is merged and `chromedp/headless-shell:latest` is updated.

Issue #1110, opened 2022-07-12
Target.domEvent panics with a concurrent frame read/write


WalkerGriggs:
#### What versions are you running?

`v7.2`

#### What did you do? Include clear steps.

```go
chromedp.ListenTarget(ctx, func(ev interface{}) {
    switch ev := ev.(type) {
        case *page.EventLifecycleEvent:
		# stop timeout and log
	case *runtime.EventConsoleAPICalled:
		# logging
	case *runtime.EventExceptionThrown:
		# more logging
	}
})
chromedp.Run(ctx)
```

#### What did you expect to see?

Logging, standard run behavior

#### What did you see instead?

```
fatal error: concurrent map read and map write

github.com/chromedp/chromedp.(*Target).domEvent
	external/com_github_chromedp_chromedp/target.go:362
```

Digging into `target.go`, there are 7 occurrences where `t.frames` is accessed. Most have a surrounding read lock

https://github.com/chromedp/chromedp/blob/03c1c7402666a6d042354c57bfb8463f704684b9/target.go#L260-L262

or write lock

https://github.com/chromedp/chromedp/blob/03c1c7402666a6d042354c57bfb8463f704684b9/target.go#L298-L305

except for `domEvent`, on line 362

https://github.com/chromedp/chromedp/blob/03c1c7402666a6d042354c57bfb8463f704684b9/target.go#L361-L364

WalkerGriggs:
I've opened #1111, but let me know if I'm missing some contribution steps

Issue #1105, opened 2022-07-06
tests pass but panic with `panic: leaked 1 temporary dirs under /tmp/chromedp-test844676543`


ZekeLu:
#### What versions are you running?

```console
$ go list -m github.com/chromedp/chromedp
v0.8.2
$ google-chrome --version
chromedp/headless-shell:103.0.5060.114
$ go version
go version go1.18.3 linux/amd64
```

#### What did you do? Include clear steps.

Run the test within the code base of chromedp: `./contrib/docker-test.sh`

#### What did you see instead?

The tests pass. But it panic with `panic: leaked 1 temporary dirs under /tmp/chromedp-test844676543`.

See https://github.com/chromedp/chromedp/runs/7207039435?check_suite_focus=true.

ZekeLu:
I think this is #592. It seems that it happens more frequently in newer chromium (starts from version 102). I will look into it later.

ZekeLu:
https://github.com/chromedp/chromedp/actions/runs/3528412137/jobs/5918487938

PR #1111, opened 2022-07-12
Guard access to Target.{frames,cur} globally

WalkerGriggs:
Some accesses to `Target.frames` and `Target.cur` are not guarded by `Target.frameMu` unintentionally. Which leads to data race.

Fixes: #1110
Fixes: #1296

### DIFF
--- a/chromedp.go
+++ b/chromedp.go
@@ -623,9 +623,9 @@ func responseAction(resp **network.Response, actions ...Action) Action {
 
 		// Obtain frameID from the target.
 		c := FromContext(ctx)
-		c.Target.frameMu.Lock()
+		c.Target.frameMu.RLock()
 		frameID = c.Target.cur
-		c.Target.frameMu.Unlock()
+		c.Target.frameMu.RUnlock()
 
 		ListenTarget(lctx, func(ev interface{}) {
 			if loaderID != "" {
--- a/chromedp.go
+++ b/chromedp.go
@@ -351,8 +351,12 @@ func (c *Context) newTarget(ctx context.Context) error {
 			if err != nil {
 				return err
 			}
+
+			c.Target.frameMu.Lock()
 			c.Target.frames[tree.Frame.ID] = tree.Frame
 			c.Target.cur = tree.Frame.ID
+			c.Target.frameMu.Unlock()
+
 			c.Target.documentUpdated(ctx)
 		}
 		return nil
--- a/target.go
+++ b/target.go
@@ -360,7 +360,10 @@ func (t *Target) pageEvent(ev interface{}) {
 
 // domEvent handles incoming DOM events.
 func (t *Target) domEvent(ctx context.Context, ev interface{}) {
+	t.frameMu.RLock()
 	f := t.frames[t.cur]
+	t.frameMu.RUnlock()
+
 	var id cdp.NodeID
 	var op nodeOp
 
--- a/target.go
+++ b/target.go
@@ -28,7 +28,7 @@ type Target struct {
 
 	messageQueue chan *cdproto.Message
 
-	// frameMu protects both frames and cur.
+	// frameMu protects frames, execContexts, and cur.
 	frameMu sync.RWMutex
 	// frames is the set of encountered frames.
 	frames       map[cdp.FrameID]*cdp.Frame


ZekeLu:
@WalkerGriggs Good catch! Thank you very much!

How did you find the issue? Is it possible to add a regress test for this issue? It's good to have one. But it's also okay to submit the fix without a regress test if it's hard/impossible to write the test.

ZekeLu:
The failed check is due to #1105, we can ignore it here.

BTW, can you change the commit message to `avoid concurrent map read in (*Target).domEvent`?

ZekeLu:
> I can spend more time digging tomorrow if that's valuable information.

Please do digging more so that we can understand the issue better. Can you run your app with the `--race` flag? It will give more details.

> Spitballing here, but could two main event loops run in parallel if Run was called multiple times given the same target?

I'm not sure what will happen in this case, but I think we should avoid doing it. Is there a use case that you need to call `Run` in parallel?

WalkerGriggs:
> Please do digging more so that we can understand the issue better.

Can do. This is part of a system that has been running for months (maybe years now), and this is the first occurrence. We've dumped the goroutines when it panicked, so maybe that'll reveal something.

> Can you run your app with the `--race` flag? It will give more details.

Maybe, but unlikely.

> I'm not sure what will happen in this case, but I think we should avoid doing it. Is there a use case that you need to call `Run` in parallel?

Oh, we avoid that; no use case here. Just throwing out some ideas, though probably not useful until we narrow things down.

WalkerGriggs:
@ZekeLu -- following up on this. We just ran into this same issue again. I would like to rebase to the latest master and merge this change in if possible.

ZekeLu:
Do you have more information to share this time?

> fatal error: concurrent map read and map write

I just took a quick look. There are only 3 map writes:

https://github.com/chromedp/chromedp/blob/1738691f6ebd5f448369e81d51aa81ff5cf6d14a/target.go#L298-L305

https://github.com/chromedp/chromedp/blob/1738691f6ebd5f448369e81d51aa81ff5cf6d14a/target.go#L345-L354

https://github.com/chromedp/chromedp/blob/1738691f6ebd5f448369e81d51aa81ff5cf6d14a/chromedp.go#L290-L298

The first 2 writes are in the `pageEvent` func. But `pageEvent` and `domEvent` won't run at the same time for the same target. So it's most likely that it is the write in `newTarget` to blame.

If this is the case, we should add a write lock here too. I think it's possible to write a test for this case. Would you like to give it a try?

BTW, the write in `newTarget` only happen when you use `chromedp.WithTargetID`. Have your used `chromedp.WithTargetID` in your project?

WalkerGriggs:
> Do you have more information to share this time?

@ZekeLu, unfortunately not enough to confidently reproduce this specific error case in tests. The stack traces are thin at best, and these rare occurrences are causing issues in production.

Locking the frame read when there's the risk of concurrent access feels like a quick win here.

> BTW, the write in newTarget only happen when you use chromedp.WithTargetID. Have your used chromedp.WithTargetID in your project?

We do use `WithTargetID` when we create a new context. This line is the top-most chromdp specific line in the stacktrace. I can probably put together the full trace dump, but might have to redact some chunks.

```
	tabCtx, _ := chromedp.NewContext(dpctx, chromedp.WithTargetID(newTarget.TargetID))
```

ZekeLu (changes requested):
We should take the changes from #1297 too.

And if you don't mind, please fix other minor issues related `frameMu` listed below:

1. it seems  that a read lock is enough:

https://github.com/chromedp/chromedp/blob/0b936bf02d9c76500cc82f0c809071df2ca97733/chromedp.go#L622-L624

2. `// frameMu protects frames, execContexts, and cur.`
https://github.com/chromedp/chromedp/blob/0b936bf02d9c76500cc82f0c809071df2ca97733/target.go#L31

WalkerGriggs:
Rebased to master (0b936bf02d9c76500cc82f0c809071df2ca97733).

@ZekeLu making those changes now

WalkerGriggs:
@ZekeLu I noticed you've triggered the tests few times which all leak a similar temp dir.

I ran the tests locally with the `./contrib/docker-test.sh` which all pass. Let me know if I can do anything to help validate the changes / get this PR over the line though -- happy to jump in

ZekeLu (approved):
Thank you @WalkerGriggs and @xiusin!

Issue #1121, opened 2022-07-20
Assignment Mismatch: 8 variables but css.GetMatchedStylesForNode...


dnapier:
Should note this is similar to #1031 & #1037 

# Error:
```bash
# github.com/chromedp/chromedp
../../../pkg/mod/github.com/chromedp/chromedp@v0.8.2/query.go:1139:33: assignment mismatch: 8 variables but css.GetMatchedStylesForNode(nodes[0].NodeID).Do returns 9 values
make: *** [run] Error 2
```

#### What versions are you running?

```console
$ go list -m github.com/chromedp/chromedp
github.com/chromedp/chromedp v0.8.2
$ google-chrome --version
Chromium 105.0.5169.0 
$ go version
go version go1.18.4 darwin/arm64
```

#### What did you do? Include clear steps.

```go
if err := chromedp.Run(ctx,
	chromedp.Navigate("http://localhost:3000/auth"),
	chromedp.SendKeys("input#username0", username),
	chromedp.SendKeys("input#password1", password),
	chromedp.Click("input[name=authorize]", chromedp.NodeEnabled), // Enter creds
	chromedp.Click("input#smsnumber0_0"),
	chromedp.Click("input#accept"),
	chromedp.Click("input#rememberdevice"), // "Trust this device..."
	chromedp.ActionFunc(func(ctx context.Context) (err error) {
		conn := models.Conn
		if _, err := conn.Exec(context.Background(), "listen smsreceive"); err != nil {
			log.Error().Err(err).Send()
		}
		code, _ := conn.WaitForNotification(context.Background())
		smscode = code.Payload
		return
	}),
	//chromedp.SendKeys("input", smscode),
	chromedp.WaitEnabled("input#authorization0"),
	chromedp.Click("input#accept", chromedp.NodeEnabled), // After put in SMS code
	chromedp.Click("input#accept", chromedp.NodeEnabled), // Scopes
```

#### What did you expect to see?
This code worked last night. Expected to see a browser window waiting for entering a SMS verification code

#### What did you see instead?
Nothing, error occurs which prevents program from running

ZekeLu:
Thanks for reporting this issue! The CDP definition was changed (again :cry: ). We will update `chromedp` to align to it later. A workaround for now is to downgrade the `github.com/chromedp/cdproto` package to `
v0.0.0-20220515234810-83d799542a04`: 

```
go get github.com/chromedp/cdproto@v0.0.0-20220515234810-83d799542a04
```

[...]

PR #1122, opened 2022-07-20
Bugfix: Include missing return value for matching struct

dnapier:
This addresses #1121

The corresponding data structure in [cdproto](https://github.com/chromedp/cdproto/blob/fd2b6dcad39e613805488dc3741d812877eab918/css/css.go#L357) was updated and broke the dependent function call in `chromedp/query.go`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/chromedp/chromedp
 go 1.18
 
 require (
-	github.com/chromedp/cdproto v0.0.0-20220515234810-83d799542a04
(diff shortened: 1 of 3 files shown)


ZekeLu:
Thank you! Please update `go.mod` too.

dnapier:
Sorry for the force-push. Squashed the `go.mod` update into a single commit.

ZekeLu:
The force-push is preferred. I should mentioned that `go.sum` should be updated too.

dnapier:
Added `go.sum` via force-push.

PR #1036, opened 2022-03-15
Allow to set custom WS URL waiting time

karelbilek:
With a heavy CPU utilization, Chrome sometimes does not start in 20 seconds.

This adds a custom option to set the WS URL waiting time.

PR #1134, opened 2022-08-09
Make chrome startup timeout configurable

8tomat8:
We are using this library to run chrome inside AWS Lambda.
We saw that sometimes it takes more than 20s to start, so this setting failed the entire job from time to time.
We set it to 120 on our prod env, and all the problems disappeared. It is running for more than a year already.

### DIFF
--- a/allocate.go
+++ b/allocate.go
@@ -117,6 +117,12 @@ type ExecAllocator struct {
 // temporary directory is used.
 var allocTempDir string
 
+// DebugURLReadTimeout defines how much time we should wait for the debug URL from Chrome after start.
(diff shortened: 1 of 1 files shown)


8tomat8:
I've been looking into better ways to pass this setting from the caller, but it might break the API.
This is quite a top-level setting, usually set once per environment, keeping it as a global var so far is the best solution I found :smiling_face_with_tear:

ZekeLu:
I think #1036 is better. But this PR provides a valid use case. Thank you!

ZekeLu:
Closing in favor of #1036.

PR #1159, opened 2022-09-13
FEATURE: make polling interval configurable

taylorchu:
5ms is hard-coded to be the default polling interval for many timers for go to check if browser condition becomes true. 5ms is often too short in some cases. For our use-case, 100ms to 1s is sufficient.

### DIFF
--- a/query.go
+++ b/query.go
@@ -555,9 +555,10 @@ func AtLeast(n int) QueryOption {
 	}
 }
 
-// RetryInterval updates query's retry interval.
+// RetryInterval is an element query action option to set the retry interval to specify
+// how often it should retry when it failed to select the target element(s).
 //
-// By default, a query retries every 5ms.
+// The default value is 5ms.
 func RetryInterval(interval time.Duration) QueryOption {
 	return func(s *Selector) {
 		s.retryInterval = interval
--- a/query.go
+++ b/query.go
@@ -7,7 +7,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/chromedp/cdproto/cdp"
 	"github.com/chromedp/cdproto/css"
@@ -155,7 +154,8 @@ func (s *Selector) Do(ctx context.Context) error {
 	if t == nil {
 		return ErrInvalidTarget
 	}
-	return retryWithSleep(ctx, 5*time.Millisecond, func(ctx context.Context) (bool, error) {
+	pollingInterval := FromContext(ctx).pollingInterval
+	return retryWithSleep(ctx, pollingInterval, func(ctx context.Context) (bool, error) {
 		frame, root, execCtx, ok := t.ensureFrame()
 		if !ok {
 			return false, nil
--- a/query.go
+++ b/query.go
@@ -42,7 +42,7 @@ type Selector struct {
 //
 // For example:
 //
-//     chromedp.Run(ctx, chromedp.SendKeys(`thing`, chromedp.ByID))
+//	chromedp.Run(ctx, chromedp.SendKeys(`thing`, chromedp.ByID))
 //
 // The above will perform a "SendKeys" action on the first element matching a
 // browser CSS query for "#thing".
@@ -51,17 +51,17 @@ type Selector struct {
 // the primary way of automating Tasks in the browser. They are typically
 // written in the following form:
 //
-//     Action(selector[, parameter1, ...parameterN][,result][, queryOptions...])
+//	Action(selector[, parameter1, ...parameterN][,result][, queryOptions...])
 //
 // Where:
 //
-//     Action         - the action to perform
-//     selector       - element query selection (typically a string), that any matching node(s) will have the action applied
-//     parameter[1-N] - parameter(s) needed for the individual action (if any)
-//     result         - pointer to a result (if any)
-//     queryOptions   - changes how queries are executed, or how nodes are waited for (see below)
+//	Action         - the action to perform
+//	selector       - element query selection (typically a string), that any matching node(s) will have the action applied
+//	parameter[1-N] - parameter(s) needed for the individual action (if any)
+//	result         - pointer to a result (if any)
+//	queryOptions   - changes how queries are executed, or how nodes are waited for (see below)
 //
-// Query Options
+// # Query Options
 //
 // By* options specify the type of element query used By the browser to perform
 // the selection query. When not specified, element queries will use BySearch
@@ -78,7 +78,7 @@ type Selector struct {
 // element query has returned one or more elements, and after the node condition is
 // true.
 //
-// By Options
+// # By Options
 //
 // The BySearch (default) option enables querying for elements by plain text,
 // CSS selector or XPath query, wrapping DOM.performSearch.
@@ -103,7 +103,7 @@ type Selector struct {
 // for querying DOM elements that cannot be retrieved using other By* funcs,
 // such as ShadowDOM elements.
 //
-// Node Options
+// # Node Options
 //
 // The NodeReady (default) option causes the query to wait until all element
 // nodes matching the selector have been retrieved from the browser.
@@ -154,8 +154,11 @@ func (s *Selector) Do(ctx context.Context) error {
 	if t == nil {
 		return ErrInvalidTarget
 	}
-	pollingInterval := FromContext(ctx).pollingInterval
-	return retryWithSleep(ctx, pollingInterval, func(ctx context.Context) (bool, error) {
+	c := FromContext(ctx)
+	if c == nil {
+		return ErrInvalidContext
+	}
+	return retryWithSleep(ctx, c.selectorPollingInterval, func(ctx context.Context) (bool, error) {
 		frame, root, execCtx, ok := t.ensureFrame()
 		if !ok {
 			return false, nil
--- a/query.go
+++ b/query.go
@@ -42,7 +42,7 @@ type Selector struct {
 //
 // For example:
 //
-//	chromedp.Run(ctx, chromedp.SendKeys(`thing`, chromedp.ByID))
+//     chromedp.Run(ctx, chromedp.SendKeys(`thing`, chromedp.ByID))
 //
 // The above will perform a "SendKeys" action on the first element matching a
 // browser CSS query for "#thing".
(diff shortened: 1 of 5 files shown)


taylorchu:
@kenshaw @ZekeLu thanks!

ZekeLu:
@taylorchu Thank you for your contribution!

I'm worried that `WithContextPollingInterval` will make the users confusing since we already have `WithPollingInterval`.

https://github.com/chromedp/chromedp/blob/d52f142d01c7ab3391991d23d957e5f9b774ee94/poll.go#L144-L150

BTW, it seems  that you are aware of the `chromedp.PollAction`. Do you have a use case that can not  be implemented with `chromedp.PollAction`?

My rule of thumb is: I will use the `chromedp.QueryAction` first because it's easy. The retry logic is to make sure it will give me the result. But if it needs several retries, I will consider using the `chromedp.PollAction.

taylorchu:
Use case: After running a few "sequential" prepare steps, our chromedp will then wait for the site to do a list of unordered things. We start one goroutine for each of these things to wait for them to happen, perform actions, and then wait for them to happen again.

We initially didn't know chromedp polls at this rapid rate, and don't want to rewrite with poll action now because it is essentially a huge refactor from go to pure js. It seems like simply making this 5ms configurable is good enough for most people.

I also think the function name could be better. If you have a suggestion, let me know.

taylorchu:
In production, we have about 10 watchdog goroutines. Polling at 1s vs 5ms improves cpu usage a lot.

ZekeLu:
> and don't want to rewrite with poll action now because it is essentially a huge refactor from go to pure js

I think the `chromedp.PollAction` + js is better since it requires less round trip. Let's assume that the project is created from scratch, do you think `WithContextPollingInterval` + go is better? And why?

taylorchu:
yes. It allows experimentation to production with the exact same code. Also the polling interval should be configurable for performance reason in general.

ZekeLu:
> It allows experimentation to production with the exact same code.

I think it holds true for the two approaches.

> Also the polling interval should be configurable for performance reason in general.

`chromedp.PollAction` has 3 polling modes:

- in requestAnimationFrame callback
- with a specified interval
- on every DOM mutation

See https://pkg.go.dev/github.com/chromedp/chromedp#hdr-Polling_Options

I want to see an example that `WithContextPollingInterval` is better than the current available solutions.

taylorchu:
> I think it holds true for both approaches.

https://github.com/chromedp/examples/blob/master/click/main.go#L27 

How do you convert this to poll action without converting go to js?

taylorchu:
The only change I made to reduce cpu usage significantly in comparison:
```
taskCtx, cancel := chromedp.NewContext(allocCtx, chromedp.WithContextPollingInterval(time.Second))
```

ZekeLu:
> How do you convert this to poll action without converting go to js?

I can't. The poll action works with js expressions. Do you mean that you don't want to use js at all?

For the record, if I have to implement it with the poll action, I will do it like this:

```diff
-		chromedp.WaitVisible(`body > footer`),
+		chromedp.Poll(`document.querySelector("body > footer") !== null`, nil, chromedp.WithPollingMutation()),
```

Notes:

1. For this page, `body > footer` is created soon after the page is loaded. Both the calls are not needed and could be removed totally.
2. `chromedp.WaitVisible("body > footer")` can be improved by using `chromedp.WaitReady("body > footer", chromedp.ByQuery)` instead.

taylorchu:
I simplified the pr, and added a test.

ZekeLu:
Thanks. Now the PR looks more reasonable. I would limit the change to the query action:

```diff
index 53a4206..f494945 100644
--- a/query.go
+++ b/query.go
@@ -29,6 +29,7 @@ type Selector struct {
    sel           interface{}
    fromNode      *cdp.Node
    exp           int
+   retryInterval time.Duration
    by            func(context.Context, *cdp.Node) ([]cdp.NodeID, error)
    wait          func(context.Context, *cdp.Frame, runtime.ExecutionContextID, ...cdp.NodeID) ([]*cdp.Node, error)
    after         func(context.Context, runtime.ExecutionContextID, ...*cdp.Node) error
@@ -130,6 +131,7 @@ func Query(sel interface{}, opts ...QueryOption) QueryAction {
    s := &Selector{
        sel:           sel,
        exp:           1,
+       retryInterval: 5 * time.Millisecond,
    }
 
    // apply options
@@ -155,7 +157,7 @@ func (s *Selector) Do(ctx context.Context) error {
    if t == nil {
        return ErrInvalidTarget
    }
-   return retryWithSleep(ctx, 5*time.Millisecond, func(ctx context.Context) (bool, error) {
+   return retryWithSleep(ctx, s.retryInterval, func(ctx context.Context) (bool, error) {
        frame, root, execCtx, ok := t.ensureFrame()
        if !ok {
            return false, nil
@@ -254,6 +256,12 @@ func (s *Selector) waitReady(check func(context.Context, runtime.ExecutionContex
    }
 }
 
+func RetryInterval(interval time.Duration) QueryOption {
+   return func(s *Selector) {
+       s.retryInterval = interval
+   }
+}
```

The caller has to pass the `chromedp.RetryInterval` option to the query action for which it wants a different retry interval. It's a little annoying, but it gives the caller finer grain control.

query.go
```diff
@@ -553,6 +555,15 @@ func AtLeast(n int) QueryOption {
 	}
 }
 
+// RetryInterval updates query's retry interval.
+//
+// By default, a query retries every 5ms.
```
ZekeLu, line 560: How about:

```
// RetryInterval is an element query action option to set the retry interval to specify
// how often it should retry when it failed to select the target element(s).
//
// The default value is 5ms.
```

query_test.go
```diff
@@ -188,6 +189,47 @@ func TestAtLeast(t *testing.T) {
...
+		ctx, cancel := context.WithTimeout(ctx, time.Second)
+		defer cancel()
+
+		if err := Run(ctx, Query("//input", WaitFunc(count), RetryInterval(10*time.Second))); err != nil && !errors.Is(err, context.DeadlineExceeded) {
+			t.Fatalf("got error: %v", err)
+		}
+		if waitCount != 1 {
+			t.Fatalf("query should retry once")
+		}
+	}
```
ZekeLu, line 230: Use sub tests instead. See

https://github.com/chromedp/chromedp/blob/d52f142d01c7ab3391991d23d957e5f9b774ee94/query_test.go#L221-L231

query_test.go
```diff
@@ -188,6 +189,47 @@ func TestAtLeast(t *testing.T) {
...
+	waitCount := 0
+	count := func(context.Context, *cdp.Frame, cdpruntime.ExecutionContextID, ...cdp.NodeID) ([]*cdp.Node, error) {
+		waitCount += 1
+		return nil, ErrHasResults
+	}
+
+	{
+		waitCount = 0
+
+		ctx, cancel := context.WithTimeout(ctx, time.Second)
```
ZekeLu, line 207: reduce the timeout. I think 100ms should be enough.

query_test.go
```diff
@@ -188,6 +189,47 @@ func TestAtLeast(t *testing.T) {
...
 }
 
+func TestRetryInterval(t *testing.T) {
+	t.Parallel()
+
+	ctx, cancel := testAllocate(t, "js.html")
+	defer cancel()
+
+	waitCount := 0
+	count := func(context.Context, *cdp.Frame, cdpruntime.ExecutionContextID, ...cdp.NodeID) ([]*cdp.Node, error) {
```
ZekeLu, line 199: Add comment to the count function

```go
// count is a wait function that makes the query always fail and count the number of retries
count := ....
```

query_test.go
```diff
@@ -188,6 +189,47 @@ func TestAtLeast(t *testing.T) {
...
+		return nil, ErrHasResults
+	}
+
+	{
+		waitCount = 0
+
+		ctx, cancel := context.WithTimeout(ctx, time.Second)
+		defer cancel()
+
+		if err := Run(ctx, Query("//input", WaitFunc(count))); err != nil && !errors.Is(err, context.DeadlineExceeded) {
```
ZekeLu, line 210: Wrap long line

query_test.go
```diff
@@ -188,6 +189,47 @@ func TestAtLeast(t *testing.T) {
...
+		}
+		if waitCount <= 1 {
+			t.Fatalf("query should retry more than once")
+		}
+	}
+
+	{
+		waitCount = 0
+
+		ctx, cancel := context.WithTimeout(ctx, time.Second)
```
ZekeLu, line 221: ditto

query_test.go
```diff
@@ -188,6 +189,47 @@ func TestAtLeast(t *testing.T) {
...
+		defer cancel()
+
+		if err := Run(ctx, Query("//input", WaitFunc(count), RetryInterval(10*time.Second))); err != nil && !errors.Is(err, context.DeadlineExceeded) {
```
ZekeLu, line 224: ditto

ZekeLu (changes requested):
Thank you! I have added some comments. Please take a look.

ZekeLu:
```go
func TestRetryInterval(t *testing.T) {
	t.Parallel()

	tests := []struct {
		name         string
		opts         []QueryOption
		wantCountMin int
		wantCountMax int
	}{
		{
			name: "default",
			opts: []QueryOption{},
			// in 100ms
			wantCountMin: 10,
			wantCountMax: 20,
		},
		{
			name: "large interval",
			opts: []QueryOption{RetryInterval(60 * time.Millisecond)},
			// in 100ms
			wantCountMin: 2,
			wantCountMax: 2,
		},
	}

	ctx, cancel := testAllocate(t, "js.html")
	defer cancel()

	for _, tc := range tests {
		t.Run(tc.name, func(t *testing.T) {
			retryCount := 0

			// count is a wait function that makes the query always fail and
			// counts the number of retries. Note that the wait func is called
			// only after the number of result nodes >= s.exp .
			count := WaitFunc(
				func(ctx context.Context, f *cdp.Frame, eci cdpruntime.ExecutionContextID, ni ...cdp.NodeID) ([]*cdp.Node, error) {
					retryCount += 1
					return nil, ErrInvalidTarget
				},
			)

			ctx, cancel := context.WithTimeout(ctx, 100*time.Millisecond)
			defer cancel()

			opts := append(tc.opts, count)
			err := Run(ctx, Query("//input", opts...))

			if err == nil || !errors.Is(err, context.DeadlineExceeded) {
				t.Fatalf("want error context.DeadlineExceeded, got: %v", err)
			}
			if retryCount < tc.wantCountMin {
				t.Fatalf("want retry count > %d, got: %d", tc.wantCountMin, retryCount)
			}
			if retryCount > tc.wantCountMax {
				t.Fatalf("want retry count < %d, got: %d", tc.wantCountMax, retryCount)
			}
		})
	}
}
```

ZekeLu (approved):

ZekeLu:
Thank you!

PR #1171, opened 2022-10-03
Add Panic handler to waitReady 

djenriquez:
Under heavy concurrency with contexts being canceled based on certain events, we were experiencing nil pointer dereference panics coming from this goroutine:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x100b422d0]

goroutine 68222 [running]:
github.com/chromedp/chromedp.parseRemoteObject(0x140026ee300?, {0x1010b05e0?, 0x14007d58550?})
        /Users/djenriquez/go/pkg/mod/github.com/chromedp/chromedp@v0.8.6/eval.go:69 +0x20
github.com/chromedp/chromedp.callFunctionOn({0x10123ae90, 0x140055faf30}, {0x100f05f24?, 0x100b3e1b8?}, {0x1010b05e0?, 0x14007d58550}, 0x14007d59320, {0x0, 0x0, 0x100b446d0?})
        /Users/djenriquez/go/pkg/mod/github.com/chromedp/chromedp@v0.8.6/call.go:68 +0x200
github.com/chromedp/chromedp.CallFunctionOn.func1({0x10123ae90?, 0x140055faf30?})
        /Users/djenriquez/go/pkg/mod/github.com/chromedp/chromedp@v0.8.6/call.go:26 +0x44
github.com/chromedp/chromedp.ActionFunc.Do(0x100f05f24?, {0x10123ae90?, 0x140055faf30?})
        /Users/djenriquez/go/pkg/mod/github.com/chromedp/chromedp@v0.8.6/chromedp.go:614 +0x34
github.com/chromedp/chromedp.callFunctionOnNode({0x10123ae90, 0x140055faf30}, 0x140064b0600, {0x100f05f24, 0x72}, {0x1010a4020, 0x14006f017f8}, {0x0, 0x0, 0x0})
        /Users/djenriquez/go/pkg/mod/github.com/chromedp/chromedp@v0.8.6/query.go:430 +0x194
github.com/chromedp/chromedp.NodeVisible.func1({0x10123ae90, 0x140055faf30}, 0x14000c9df88?, 0x140064b0600)
        /Users/djenriquez/go/pkg/mod/github.com/chromedp/chromedp@v0.8.6/query.go:460 +0x158
github.com/chromedp/chromedp.(*Selector).waitReady.func1.1(0x14000c9dfa8?)
        /Users/djenriquez/go/pkg/mod/github.com/chromedp/chromedp@v0.8.6/query.go:239 +0x78
created by github.com/chromedp/chromedp.(*Selector).waitReady.func1
        /Users/djenriquez/go/pkg/mod/github.com/chromedp/chromedp@v0.8.6/query.go:235 +0x250
```

When I dug into this, the panic occurs when a `*runtime.RemoteObject` parameter returns a nil, here: https://github.com/chromedp/chromedp/blob/master/eval.go#L69. Digging further, this seems to happen because the goroutine executing `check()` depends on a context that _can_ be canceled in a different goroutine. I believe this leads to the `*runtime.RemoteObject` being nil, and then triggering the panic when its `.Type` property is being accessed.

I'm not sure if this is the best method, as this sort of goes around the problem, but I'm not sure how one can guarantee that the goroutine doesn't panic when it depends on a context defined outside its scope.

### DIFF
--- a/query.go
+++ b/query.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"runtime/debug"
 	"strconv"
 	"strings"
 	"sync"
@@ -44,7 +45,7 @@ type Selector struct {
 //
 // For example:
 //
-//     chromedp.Run(ctx, chromedp.SendKeys(`thing`, chromedp.ByID))
+//	chromedp.Run(ctx, chromedp.SendKeys(`thing`, chromedp.ByID))
 //
 // The above will perform a "SendKeys" action on the first element matching a
 // browser CSS query for "#thing".
@@ -53,17 +54,17 @@ type Selector struct {
 // the primary way of automating Tasks in the browser. They are typically
 // written in the following form:
 //
-//     Action(selector[, parameter1, ...parameterN][,result][, queryOptions...])
+//	Action(selector[, parameter1, ...parameterN][,result][, queryOptions...])
 //
 // Where:
 //
-//     Action         - the action to perform
-//     selector       - element query selection (typically a string), that any matching node(s) will have the action applied
-//     parameter[1-N] - parameter(s) needed for the individual action (if any)
-//     result         - pointer to a result (if any)
-//     queryOptions   - changes how queries are executed, or how nodes are waited for (see below)
+//	Action         - the action to perform
+//	selector       - element query selection (typically a string), that any matching node(s) will have the action applied
+//	parameter[1-N] - parameter(s) needed for the individual action (if any)
+//	result         - pointer to a result (if any)
+//	queryOptions   - changes how queries are executed, or how nodes are waited for (see below)
 //
-// Query Options
+// # Query Options
 //
 // By* options specify the type of element query used By the browser to perform
 // the selection query. When not specified, element queries will use BySearch
@@ -80,7 +81,7 @@ type Selector struct {
 // element query has returned one or more elements, and after the node condition is
 // true.
 //
-// By Options
+// # By Options
 //
 // The BySearch (default) option enables querying for elements by plain text,
 // CSS selector or XPath query, wrapping DOM.performSearch.
@@ -105,7 +106,7 @@ type Selector struct {
 // for querying DOM elements that cannot be retrieved using other By* funcs,
 // such as ShadowDOM elements.
 //
-// Node Options
+// # Node Options
 //
 // The NodeReady (default) option causes the query to wait until all element
 // nodes matching the selector have been retrieved from the browser.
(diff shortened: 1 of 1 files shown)


ZekeLu:
Thank you @djenriquez !

I think we should find out why `*runtime.RemoteObject` is nil first and handle it directly (instead of recovering from a panic).

> this seems to happen because the goroutine executing check() depends on a context that can be canceled in a different goroutine. I believe this leads to the `*runtime.RemoteObject` being nil

This is unlikely. If the context is cancelled, if should return before reaching `parseRemoteObject`:

https://github.com/chromedp/chromedp/blob/6a8fef8d17ceaf7dd5d64ac614f207c9b7e9d15a/call.go#L60-L68

Can you turn on the debug log (`chromedp.WithDebugf(log.Printf)`) and share the log? We are interested in CDP messages like the one shown below. But please share more logs around these messages.

```
2022/10/03 09:32:37 -> {"id":21,"sessionId":"8743C20757D14133E1016B3BC85C9CB8","method":"Runtime.callFunctionOn","params":{"functionDeclaration":"function visible() {\n    return Boolean(this.offsetWidth || this.offsetHeight || this.getClientRects().length);\n}\n","objectId":"6922537476352520887.2.1","silent":true,"returnByValue":true}}
2022/10/03 09:32:37 <- {"id":21,"result":{"result":{"type":"boolean","value":true}},"sessionId":"8743C20757D14133E1016B3BC85C9CB8"}
```

djenriquez:
Definitely, I'll enable those logs and get them over as soon as I can. 

In the meantime, to give context, we monitor specific network requests and send specific events to an event handler to do some automation. Each event generates its own context with a cancel function to use for the page automation. Sometimes a page load will send the same event, in which case, we cancel any existing contexts that are being processed so that we don't overlap any automation that hasn't finished. It's during this cancel where the panic occurs.

ZekeLu:
> [signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x100b422d0]

Pay attention to the code `0x2`, which is `SEGV_ACCERR`. Maybe `v` is not nil. See https://stackoverflow.com/a/54767503/1369400.

BTW, can you run your app with the `-race` option? See https://go.dev/doc/articles/race_detector

djenriquez:
@ZekeLu Thanks so much for pointing that out. Turns out my mind was used to ignoring that first part of the error and lazer focusing on the nil pointer defer. Was able to find the issue in our implementation and corrected it, verified no issues afterwards.

I don't think this PR is needed anymore as it _would_ hide problems like it did with this forked code.

Thanks for your input!

ZekeLu:
Thank you for your feedback! I'm curious about how this could happen. Please share more about the error if it's okay with you.

PR #990, opened 2022-01-11
Add option to NewRemoteAllocator to avoid URL detection

andyMrtnzP:
# Summary

This PR solves issue https://github.com/chromedp/chromedp/issues/972. `RemoteAllocatorOption` was added, which contains `FetchJSONDebuggerURL`. This could avoid URL detection on `NewRemoteAllocator`. `PersistDebuggerURL`, an instance of `RemoteAllocatorOption` was added, with `FetchJSONDebuggerURL` as false.

## Example

Consider connecting to a `wss://` URL

``` go
func main() {
  var devToolWsUrl string
  var title string
  var opts = chromedp.PersistDebuggerURL
  var url = "wss://chrome.browserless.io?token=my_token_here"

  flag.StringVar(&devToolWsUrl, "devtools-ws-url", url, "DevTools Websocket URL")
  flag.Parse()

  actxt, cancelActxt := chromedp.NewRemoteAllocator(context.Background(), devToolWsUrl, opts)
  // works as well      chromedp.NewRemoteAllocator(context.Background(), devToolWsUrl, chromedp.RemoteAllocatorOption{ FetchJSONDebuggerURL: false })
  defer cancelActxt()

  ctxt, cancelCtxt := chromedp.NewContext(actxt) // create new tab
  defer cancelCtxt()                             // close tab afterwards

  if err := chromedp.Run(ctxt,
    chromedp.Navigate("https://example.com"),
    chromedp.Title(&title),
  ); err != nil {
    log.Fatalf("Failed getting title of example.com: %v", err)
  }

  log.Println("Got title: `" + title + "`")
}
```
will connect to the WebSocket and scrape the page correctly -- it currently does not connect and errs a `HTTP 301`

PR #1177, opened 2022-10-13
remote allocator support option

GanymedeNil:
## Changes
- Added `RemoteAllocatorOption` parameter to `NewRemoteAllocator`.
- `Param` is a generic option to pass a flag to remote url. If the value is a string and isFlag is true, it will be passed as --name=value. If it's a boolean, it will be passed as --name if value is true. If isFlag is false, it will be passed as name=value. If it's a boolean, it will be passed as name if value is true.

## Demo
```go
	devtoolsWsURL := "wss://chrome.browserless.io"
	opts := []chromedp.RemoteAllocatorOption{
		chromedp.Param("token", "YOUR_API_KEY", false),
		chromedp.Param("user-agent", "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/100.0.4896.127 Safari/537.36", true),
	}
	allocatorContext, cancel := chromedp.NewRemoteAllocator(context.Background(), devtoolsWsURL, opts...)
```

### DIFF
--- a/allocate.go
+++ b/allocate.go
@@ -12,6 +12,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


ZekeLu:
@GanymedeNil Thank you!

Please note that URI components can not be constructed like this: `fmt.Sprintf("%s=%s", name, value)`. They should be escaped.

Besides that, it seems that what the PR does is to construct a special URL for browserless, which I think don't have to be a part of `chromedp`. We are aware of that the current `chromedp` does not work with `browserless`. #990 attempts to address this issue. I want to address the issue in another way. I hope that I can prepare a PR in the next few days, and I hope that you can give it a try when it's ready.

Thank you again!

ZekeLu:
@GanymedeNil Sorry that I forgot to @ you from #1184. Now that #1184 was merged, I'm going to close this one. Please let me know if you think this one should not be closed.

PR #1184, opened 2022-10-26
introduce a remote allocator option: NoModifyURL

### LINKED ISSUE #972: Avoid modifying WSS URLs since parameters are dropped through the detectURL method
### Bug description
Ever since the chromedp v0.7.3, the **detectURL** method was implemented in **util.go** to make it easier to send the base URL of a WS to NewRemoteAllocator. The rest of the URL is built by fetching the **webSocketDebuggerUrl** from `json/version` and constructing the URL again; however if the websocket is passed on with parameters, such as the case of an authentication token to a WSS for a hosted browser service, then it is impacting connection, since the parameter is dropped and the connection isn't successful.

Steps to reproduce the problem:

1. Use chromedp v0.7.3 or above
2. Pass a WSS to the NewRemoteAllocator

This throws the following error
`2021/12/03 14:42:06 Failed getting title of example.com: could not dial "ws://chrome.browserless.io": unexpected HTTP response status: 301 `

```console
github.com/chromedp/chromedp v0.7.3
$ google-chrome --version
Google Chrome 96.0.4664.55
$ go version
go version go1.17.3 darwin/amd64
```

#### What was run to get this error message?

Run the Go script below
```go
package main

import (
  "context"
  "flag"
  "log"

)

func main() {
  var devToolWsUrl string
  var title string

  flag.StringVar(&devToolWsUrl, "devtools-ws-url", "wss://chrome.browserless.io?token=my_token_here", "DevTools Websocket URL")
  flag.Parse()

  actxt, cancelActxt := chromedp.NewRemoteAllocator(context.Background(), devToolWsUrl)
  defer cancelActxt()

  ctxt, cancelCtxt := chromedp.NewContext(actxt) // create new tab
  defer cancelCtxt()                             // close tab afterwards

  if err := chromedp.Run(ctxt,
    chromedp.Navigate("https://example.com"),
    chromedp.Title(&title),
  ); err != nil {
    log.Fatalf("Failed getting title of example.com: %v", err)
  }

  log.Println("Got title of:", title)
}
```

I was expecting the WebSocket to connect properly and scrape the Title, which occurs when using v0.7.2.
```console
$ go get -u github.com/chromedp/chromedp@v0.7.2
$ go run chrome.go                             
2021/12/03 14:42:35 Got title of: Example Domain
```

I saw the following error message, with an unexpected HTTP response due to the fact that the token was not provided when the URL was reformatted through the detectURL method in util.go
```console
$ go get -u github.com/chromedp/chromedp@v0.7.3
$ go run chrome.go
2021/12/03 14:42:06 Failed getting title of example.com: could not dial "ws://chrome.browserless.io": unexpected HTTP response status: 301
exit status 1

```
### Possible issue
This most likely has to do with PR #817 
While trying to simplify the way the URL is passed on to NewRemoteAllocator, it is unfortunately removing the arguments passed after the hostname at the moment it reconstructs the URL with the `/json/version` payload. Important arguments are lost in the process such as the token that is used to authenticate to many hosted browser services. 

### Possible fixes and justification
If the detectURL method returns Secure WebSockets as is, then they won't be modified by this function.
Why would this be important for WSS? Because most hosted browser services are running on WSS and are mainly the ones that require to authenticate through a token, which sometimes are passed through an argument such as this case and would make it important for it not be dropped from the URL.

**_Possible fixes_**

1) **Ignore urls containing "/devtools/browser/" and also the ones containing "wss"** since these endpoints could use params as authentication methods to hosted services. Trusting that WSS will most likely be hosted services and are probably going to point exactly to the correct path anyways, and would not require the assistance that the PR #817 provides.
Something like this could work for the util.go:
```go
func detectURL(urlstr string) string {
	if strings.Contains(urlstr, "/devtools/browser/") || strings.HasPrefix(urlstr, "wss") {
		return urlstr
	}
	// replace the scheme and path to construct the URL like:
	// http://127.0.0.1:9222/json/version
	u, err := url.Parse(urlstr) 
```

2) **Modify the detectURL method to extract the arguments provided after the hostname, so that they can be appended** to the webSocketDebuggerUrl that is retrieved from `json/version`.

ZekeLu:
The feature implemented by #817 makes chromedp not work with [browserless](https://www.browserless.io/). In order to address this issue, a new remote allocator option is introduced: `chromedp.NoModifyURL`. Usage:

```go
ctx, cancel := chromedp.NewRemoteAllocator(context.Background(), "wss://chrome.browserless.io?token=YOUR-API-TOKEN", chromedp.NoModifyURL)
```

Fixes #972.

### DIFF
--- a/allocate.go
+++ b/allocate.go
@@ -95,7 +95,7 @@ func NewExecAllocator(parent context.Context, opts ...ExecAllocatorOption) (cont
 	return ctx, cancelWait
 }
 
-// ExecAllocatorOption is a exec allocator option.
+// ExecAllocatorOption is an exec allocator option.
 type ExecAllocatorOption = func(*ExecAllocator)
 
 // ExecAllocator is an Allocator which starts new browser processes on the host
@@ -510,6 +510,7 @@ func WSURLReadTimeout(t time.Duration) ExecAllocatorOption {
 // NewRemoteAllocator creates a new context set up with a RemoteAllocator,
 // suitable for use with NewContext. The url should point to the browser's
 // websocket address, such as "ws://127.0.0.1:$PORT/devtools/browser/...".
+//
 // If the url does not contain "/devtools/browser/", it will try to detect
 // the correct one by sending a request to "http://$HOST:$PORT/json/version".
 //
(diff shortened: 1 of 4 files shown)


ZekeLu:
@andyMrtnzP (from #990) and @AlexLoyola @joelgriffith @kotchuprik (from #972)

Sorry for the belated fix! Would you like to review this PR and give it a try? The commits are better to be reviewed one by one because each commit just focuses on one thing and is relatively small.

Thank you!

andyMrtnzP:
Works like a charm!

ZekeLu:
Thank you for the feedback! I'm going to merge this PR.

PR #1221, opened 2022-12-12
update the screenshot funcs to match the behavior of chrome commands

### LINKED ISSUE #1220: Cepture screenshot returns full height screenshot instead of viewport height
Hey! I'm using chromedp to capture the website's screenshot for the given viewport. But when I call `chromedp.CaptureScreenshot(&buf)`, full height screenshot is taken, instead of just the viewport height.

```console
github.com/chromedp/chromedp v0.8.6
$ google-chrome --version
Chromium 108.0.5359.94 built on Debian 11.5, running on Debian 11.5
$ go version
go version go1.18.9 linux/arm64
```

```go
package crawler

import (
    "context"
    "log"
    "time"
)

type Crawler struct {
    URL            string
    ViewportWidth  int64
    ViewportHeight int64
}

func (c *Crawler) GetScreenshot() (error, []byte) {
    ctx, cancel := chromedp.NewContext(
        context.Background(),
    )
    defer cancel()

    ctx, cancel = context.WithTimeout(ctx, 10*time.Second)
    defer cancel()

    var buf []byte

    println(c.ViewportWidth, c.ViewportHeight)

    if err := chromedp.Run(ctx,
        chromedp.EmulateViewport(c.ViewportWidth, c.ViewportHeight),
        chromedp.Navigate(c.URL),
        chromedp.CaptureScreenshot(&buf),
    ); err != nil {
        log.Println(err)
        return err, nil
    }

    return nil, buf
}
```

You can also check the source code here: https://github.com/marketron-app/image-engine/blob/main/crawler/crawler.go

Get a screenshot with height of a viewport.

Got a screenshot with full height of the website.

ZekeLu:
Fixes #1220.

### DIFF
--- a/screenshot.go
+++ b/screenshot.go
@@ -59,6 +59,7 @@ func Screenshot(sel interface{}, picbuf *[]byte, opts ...QueryOption) QueryActio
 		buf, err := page.CaptureScreenshot().
 			WithFormat(page.CaptureScreenshotFormatPng).
 			WithCaptureBeyondViewport(true).
+			WithFromSurface(true).
 			WithClip(&clip).
 			Do(ctx)
 		if err != nil {
@@ -88,7 +89,7 @@ func CaptureScreenshot(res *[]byte) Action {
 	return ActionFunc(func(ctx context.Context) error {
 		var err error
 		*res, err = page.CaptureScreenshot().
-			WithCaptureBeyondViewport(true).
+			WithFromSurface(true).
 			Do(ctx)
 		return err
 	})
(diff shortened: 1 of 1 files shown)


ZekeLu:
Here are the chrome commands:

image

It seems that  the behavior changed in Chrome 108.

ZekeLu:
The test failed because it tests with `chromedp/headless-shell:latest`,  which is `chromedp/headless-shell:107.0.5304.107` as of now.

PR #1222, opened 2022-12-12
add exposefunc feature

hillguo:
In some scenarios, need to bind go func to the browser environment. when js called, the function executes in go process and returns a Promise
refer  https://pptr.dev/api/puppeteer.page.exposefunction/


ZekeLu:
Thank you! I will take a careful look later.

Can you add some tests for the new feature?

expose.go
```diff
@@ -0,0 +1,160 @@
...
+		_, err := page.AddScriptToEvaluateOnNewDocument(expression).Do(ctx)
+		return err
+	}))
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+const exposedFunJS = `
```
ZekeLu, line 109: Extrat the js code into a standalone file and save it to the `js` folder. Then load the js code with `//go:embed`.

expose.go
```diff
@@ -0,0 +1,160 @@
...
+	const callCDP = self[name];
+	console.log("callCDP",callCDP)
+	// We replace the CDP binding with a Puppeteer binding.
+	Object.assign(self, {
+		[name](args) {
+			if(typeof args != "string"){
+				return Promise.reject(new Error('function takes exactly one argument, this argument should be string'))
+			}
+			var _a, _b;
+			// This is the Puppeteer binding.
```
ZekeLu, line 133: It looks confusing to see Puppeteer in the js code.

expose.go
```diff
@@ -0,0 +1,160 @@
...
+			return new Promise((resolve, reject) => {
+				callPuppeteer.callbacks.set(seq, { resolve, reject });
+			});
+		},
+	});
+}
+`
+
+func deliverError(name string, seq int64, message, stack string) string {
+	var cmd string = `deliverError("%s",%d,"%s","%s");`
```
ZekeLu, line 148: I think it's not safe to construct js code like this. Maybe `chromedp.CallFunctionOn` can be used here?

ZekeLu (commented):
I will take a careful look later.

expose.go
hillguo: this js code is generated by Puppeteer's ts code. i will remove it

expose.go
hillguo: ok. I didn't know this function existed before.

hillguo: it has been modified to CallFunctionOn. 
By the way, I'd like to know how to use chromedp.CallFunctionOn without executionContextId or objectId?

ZekeLu: According to [Runtime.callFunctionOn](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#method-callFunctionOn):

> Either objectId or executionContextId should be specified.

expose.go
```diff
@@ -0,0 +1,117 @@
...
+func ExposeFunc(ctx context.Context, fnName string, fn BindingFunc) error {
+	c := FromContext(ctx)
+	if c == nil {
+		return ErrInvalidContext
+	}
+	if c.Target == nil {
+		if err := c.newTarget(ctx); err != nil {
+			return err
+		}
+	}
```
ZekeLu, line 43: If `ExposeFunc` is an Action, this part can be removed.

expose.go
```diff
@@ -0,0 +1,117 @@
...
+
+// BindingFunc expose function type
+type BindingFunc func(args string) (string, error)
+
+// ExposeFunc The method adds a function called name on the browser page's window object.
+// When called, the function executes BindingFunc in go env
+// and returns a Promise which resolves to the return value of BindingFunc.
+// Note. compared with puppeteer's exposeFunction.
+// the BindingFunc takes exactly one argument, this argument should be string
+func ExposeFunc(ctx context.Context, fnName string, fn BindingFunc) error {
```
ZekeLu, line 34: Can we implement `ExposeFunc` as an Action so that it can be run with  `chromedp.Run`?

expose.go
```diff
@@ -0,0 +1,117 @@
...
+	if c == nil {
+		return ErrInvalidContext
+	}
+	if c.Target == nil {
+		if err := c.newTarget(ctx); err != nil {
+			return err
+		}
+	}
+
+	c.Target.bindingFuncListenOnce.Do(func() {
```
ZekeLu, line 45: If `ExposeFunc` is an Action, is it ok to remove `bindingFuncListenOnce.Do`? Please confirm whether `addTargetBinding` can be defined with `page.AddScriptToEvaluateOnNewDocument` multiple times.

expose.go
```diff
@@ -0,0 +1,117 @@
...
+					return
+				}
+
+				c.Target.bindingFuncMu.RLock()
+				defer c.Target.bindingFuncMu.RUnlock()
+
+				result := "bindingCall name not exsit"
+				callFnName := deliverError
+
+				if fn, ok := c.Target.bindingFuncs[payload.Name]; ok {
```
ZekeLu, line 76: Can we use `fn` from the parameter of the `ExposeFunc` directly? If yes, then we don't need `bindingFuncs` (that means we won't check duplicated exposed func names, thus  remove `ErrExposeNameExist` too).

js/expose.js
```diff
@@ -0,0 +1,35 @@
...
+			(_a = callChromedp.callbacks) !== null && _a !== void 0 ? _a : (callChromedp.callbacks = new Map());
+			const seq = ((_b = callChromedp.lastSeq) !== null && _b !== void 0 ? _b : 0) + 1;
+			callChromedp.lastSeq = seq;
+			callCDP(JSON.stringify({ type, name, seq, args }));
+			return new Promise((resolve, reject) => {
+				callChromedp.callbacks.set(seq, { resolve, reject });
+			});
+		},
+	});
+}
```
ZekeLu, line 35: Add trailing new line.

js/expose.js
```diff
@@ -0,0 +1,35 @@
...
+	window[name].callbacks.get(seq).reject(error);
+	window[name].callbacks.delete(seq);
+}
+
+function deliverResult(name, seq, result) {
+	window[name].callbacks.get(seq).resolve(result);
+	window[name].callbacks.delete(seq);
+}
+
+function addTargetBinding(type, name) {
```
ZekeLu, line 13: Will it cause an unexpected result if the window has a function with the same name? If yes, maybe we should choose a special name that will reduce the risk of conflict.

js/expose.js
```diff
@@ -0,0 +1,35 @@
...
+	const error = new Error(message);
+	error.stack = stack;
+	window[name].callbacks.get(seq).reject(error);
+	window[name].callbacks.delete(seq);
+}
+
+function deliverResult(name, seq, result) {
+	window[name].callbacks.get(seq).resolve(result);
+	window[name].callbacks.delete(seq);
+}
```
ZekeLu, line 11: Maybe these two functions have the same risk of name conflict. `chromedp.CallFunctionOn` accepts function declaration. So you can avoid defining them with `page.AddScriptToEvaluateOnNewDocument`.

expose_test.go
```diff
@@ -0,0 +1,71 @@
...
+			result: "param1param1",
+		},
+		{
+			name: "func2",
+			Func: func(param string) (string, error) {
+				return strings.ToUpper(param), nil
+			},
+			param:  "param2",
+			result: "PARAM2",
+		},
```
ZekeLu, line 35: puppeteer's examples explain the purpose of `exposeFunc` better. Maybe we can copy them as test cases to chromedp. https://github.com/puppeteer/puppeteer/blob/1668d47b2f4d6176438ec1dd3165015181a6fb77/packages/puppeteer-core/src/api/Page.ts#L1160-L1210

ZekeLu (commented):
This is the second round. Please take a look. Thank you!

expose.go
hillguo: ExposeFunc It's not like it can be executed multiple times。

expose.go
hillguo: By modifying the code, it can be executed multiple times, but the effect is equivalent to function coverage

expose_test.go
hillguo: Thanks for reading and advice。

I think exposeFunc is more like a js bridge implementation，It needs to be injected into the js environment before navigation to ensure that js can be called at any time.  As a function it exists once is enough.

Maybe we need a deeper conversation to make him more reasonable

hillguo:
According to your suggestion, the implementation method has been modified. 
1. as an action to call
2. remove the lock
3. modify test code

Please review  it again

ZekeLu:
I'm sorry that it takes  a little long to review. I just want to let you know that I'm still working on it.

I'm confusing by this feature. IIUC, `runtime.AddBinding` adds a new js function to the page. It does not replace an existing function in the page. That means this feature is only useful when the web page is written with the mind that the page will be accessed by a browser that is controlled a Go application, which will expose a js function to be called by the web page. I can not think of such kind of use cases.

Do you have a real world use case to share? Thank you!

hillguo:
@ZekeLu Yes, I think so, too. 
There are two scenarios: 
1. The browser knows in advance that there is a callable function
2. After chromedp exposes the function, chromedp initiatively calls

Scenario 1, in order to ensure that the function must exist, it usually needs to be exposed before the page js is loaded. (AddScriptToEvaluateOnNewDocument) does this, but it needs to take effect after navigation. 

My business scenario is cloud Web game rendering. 
A brief introduction to the exposed functions:

1. **GameLoaded**. After the game is loaded, js calls this function, and when chromedp receives it, it calls the logic to capture the game screen. 
2. **HeartbeatReport**. Monitor the normal operation of the game. The reason why we expose go heartbeat function to browser is that heartbeat service in the private network does not have https, but we do not want to use the public network https. 
3. **VideoSave**. The data recorded by the browser is saved locally.

The two most important points for this feature: 
1. Js environment call 
2. go environment execution

ZekeLu:
Thank you for the explanation!

I'm not sure I understand it. But it seems that in the use cases, the js environment does not need the returned value from the Go environment. If this is the case, `runtime.AddBinding` is enough.

hillguo:
In my scenario, I don't need a return value and `runtime.AddBinding` is enough.
Other scenes For example:
1. you need a complex encryption algorithm function without js implementation or js running is inefficient; 
2. read local text content 

I'm sorry that my example is not very accurate , more examples can search for puppeteer-related usage 
 [exposefunction](https://pptr.dev/api/puppeteer.page.exposefunction) 

Finally, I think this feature is similar to JSBridge, a means of communication, but not necessary.

ZekeLu:
I have put the suggested changes into this commit: https://github.com/ZekeLu/chromedp/commit/feef44ecf61a01b467fe20bde2963dde3814ee0e. The tests could be improved too, but I haven't touched it yet.

Please take a look. If you agree with the suggested changes, please apply them to your commit.

Please note that I'm still not sure whether this function is useful enough to be shipped by `chromedp`.

hillguo:
Thank you for the code optimization.  It's more elegant.😊

ZekeLu (changes requested):
Thank you for your hard work too!

I have added some more comments.

I think some of them are hard to address.

I'm inclined to convert this feature to an example and add it to https://github.com/chromedp/examples instead. If we provide it as an example, we don't have to address the reminding issues. And we can add it back to the package if someone find it useful and there are attracting use cases. What do you think?

expose.go
```diff
@@ -0,0 +1,111 @@
...
+type ExposeAction Action
+
+// Expose is an action to add a function called fnName on the browser page's
+// window object. When called, the function executes fn in the Go env and
+// returns a Promise which resolves to the return value of fn.
+//
+// Note:
+// 1. This is the lite version of puppeteer's [page.exposeFunction].
+// 2. It adds "chromedpExposeFunc" to the page's window object too.
+// 3. The exposed function survives page navigation until the tab is closed?
```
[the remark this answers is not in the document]
hillguo: yes, no way to revert this action.  
```
`Runtime.removeBinding` does not remove binding function from global object 
but unsubscribes current runtime agent from Runtime.bindingCalled notifications.
```

expose.go
```diff
@@ -0,0 +1,111 @@
...
+// 4. (iframe?)
```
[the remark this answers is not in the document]
hillguo: No support，I'll add it right away.

js/expose.js
```diff
@@ -0,0 +1,43 @@
+var chromedpExposeFunc = chromedpExposeFunc || {
+  bindings: {},
+  deliverError: function (name, seq, message) {
+    const error = new Error(message);
+    chromedpExposeFunc.bindings[name].callbacks.get(seq).reject(error);
+    chromedpExposeFunc.bindings[name].callbacks.delete(seq);
+  },
+  deliverResult: function (name, seq, result) {
+    chromedpExposeFunc.bindings[name].callbacks.get(seq).resolve(result);
```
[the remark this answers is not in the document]
hillguo: 1. I'm not sure that `CallFunctionOn`  param `ExecutionContextID` can or not  avoid this problem. 
2. I agree to check the existence of seq to avoid js exceptions.
3. Is UUID too complex? Puppeteer doesn't handle this situation.  The callback is also specified to  `event.executionContextId`.
  ```
Puppeteer #onBindingCalled

this.#client
      .send('Runtime.evaluate', {
        expression,
        contextId: event.executionContextId,
      })
      .catch(debugError);
```

File: expose.go [hunk 15]
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func addScriptToEvaluateOnNewDocument(script string) Action {
+	return ActionFunc(func(ctx context.Context) error {
+		_, err := page.AddScriptToEvaluateOnNewDocument(script).Do(ctx)
+		return err
+	})
+}
+
+func evaluateOnAllFrames(script string) Action {
```
hillguo, line 114: @ZekeLu expose the function to all frames. please check this function.

js/expose.js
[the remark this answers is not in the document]
ZekeLu: You're correct. `ExecutionContextID` can prevent this problem. My bad. So we don't need UUID (BTW, when I wrote UUID, I just meant a unique id to some extend. For example, `Math.random()` should be enough).

expose.go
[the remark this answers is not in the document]
ZekeLu: Not only the binding, but also the scripts that will be evaluated on new document (`page.AddScriptToEvaluateOnNewDocument`). I know that we can store the `ScriptIdentifier` and remove it later, but it will make the code complicated.

expose.go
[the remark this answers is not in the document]
ZekeLu: Sorry, I should make it clear that I don't ask to expose it to all frames. I just want to say that we should mention the difference.

In fact, I like the idea that it's only exposed to a single frame. But if we only expose it to a single frame, then we should find a way to specify a target frame other than the root document (something like `chromedp.FromNode`). That will make things complicated. So I think exposing to all frames is acceptable.

Thanks!

expose.go
ZekeLu: We need a test case to cover it.

expose.go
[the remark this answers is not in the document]
hillguo: It only takes effect on the page after navigation. The target(Page) hasn't changed. this kind of behavior should be understood

expose.go
hillguo: Test cases have been added and a bug has been fixed

expose.go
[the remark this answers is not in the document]
ZekeLu: Most state does not survive page navigation. But the scripts that the expose action uses survive page navigation and there is no way to stop that (unless the tab is closed), so this behavior should be documented.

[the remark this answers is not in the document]
hillguo: I get it.

expose.go
```diff
@@ -0,0 +1,130 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/chromedp/cdproto/page"
+	"github.com/chromedp/cdproto/runtime"
+)
+
+// ExposedFunc is the function type that can be exposed to the browser env.
+type ExposedFunc func(args string) (string, error)
```
lindowx, line 13: I think it would be more flexible if the args parameter was defined as an interface{} type.

hillguo: The reason for this design is that it is consistent with the cdp bindingCalled event parameter, which can only pass string.
Refer to ``Runtime.bindingCalled`` 
https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#event-bindingCalled
```
PARAMETERS
    name string
    payload string

var payload struct {
    Type string `json:"type"`
    Name string `json:"name"`
    Seq  int64  `json:"seq"`
    Args string `json:"args"`
}```

PR #1227, opened 2022-12-26
Add test if downloaded file exists after download

dithmer:
In this pull request, there is simply logic added to improve testing of downloading files into a directory.

### DIFF
--- a/chromedp_test.go
+++ b/chromedp_test.go
@@ -650,11 +650,13 @@ func TestDownloadIntoDir(t *testing.T) {
 
 	dir := t.TempDir()
 
+	downloadContents := "some binary data"
+
 	s := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch r.URL.Path {
 		case "/data.bin":
 			w.Header().Set("Content-Type", "application/octet-stream")
-			fmt.Fprintf(w, "some binary data")
+			fmt.Fprintf(w, downloadContents)
 		default:
 			w.Header().Set("Content-Type", "text/html")
 			fmt.Fprintf(w, `go <a id="download" href="/data.bin">download</a> stuff/`)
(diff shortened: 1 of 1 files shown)


chromedp_test.go
```diff
@@ -642,30 +643,68 @@ func TestDownloadIntoDir(t *testing.T) {
 	ctx, cancel := testAllocate(t, "")
 	defer cancel()
 
+	// Wrap the context into a timeouted context, to ensure test with not run endless if download fails
+	// A second should be enough for this mocked download to finish, even on slower machines
+	ctx, cancel = context.WithTimeout(ctx, 1*time.Second)
+	defer cancel()
+
```
ZekeLu, line 650: Let's remove this part. Let the go test tool handle test timeout.

chromedp_test.go
```diff
@@ -642,30 +643,68 @@ func TestDownloadIntoDir(t *testing.T) {
...
+		}
+
+		fileContents, err := os.ReadFile(path)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if string(fileContents) != downloadContents {
+			t.Fatalf("Expected content of file to be '%s', but got '%s'", downloadContents, string(fileContents))
+		}
```
ZekeLu, line 706: I think we could trust that the browser will save the file correctly. To simplify the test code, let's remove the checking of the file content.

ZekeLu (changes requested):
Thank you!

chromedp_test.go
dithmer: Fully agree, changed that!

chromedp_test.go
dithmer: Alright, didn't know, this is something integrated into the tooling!

dithmer:
Thank you for the feedback, I changed your hints. Anything else I can do here?

chromedp_test.go
```diff
@@ -656,16 +657,38 @@ func TestDownloadIntoDir(t *testing.T) {
...
 		Click("#download", ByQuery),
 	); err != nil {
 		t.Fatal(err)
 	}
 
-	// TODO: wait for the download to finish, and check that the file is in
-	// the directory.
+	select {
+	case <-ctx.Done():
+		t.Fatal("chromedp context is done, before download was finished")
```
ZekeLu, line 680: Replace this line with:

```go
t.Fatalf("unexpected error: %v", ctx.Err())
```

chromedp_test.go
```diff
@@ -656,16 +657,38 @@ func TestDownloadIntoDir(t *testing.T) {
...
+	case guid := <-done:
+		path := filepath.Join(dir, guid)
+
+		if _, err := os.Stat(path); err != nil {
+			if errors.Is(err, os.ErrNotExist) {
+				t.Fatal("Downloaded file does not exist")
+			}
+
+			t.Fatal("Unknown error while checking downloaded file")
+		}
```
ZekeLu, line 690: Replace the lines with:

```go
if _, err := os.Stat(filepath.Join(dir, guid)); err != nil {
	t.Fatal("want error nil, got: %v", err)
}
```

ZekeLu (changes requested):
Thanks!

ZekeLu:
Merged.

Issue #1259, opened 2023-02-13
CallFunctionOn does not properly handle the null returned by js


xueyc1f:
#### What versions are you running?

```console
$ go list -m github.com/chromedp/chromedp
github.com/chromedp/chromedp v0.8.6
$ google-chrome --version
109.0.5414.120
$ go version
go version go1.19 windows/amd64
```

#### What did you do? Include clear steps.

```go
package main

import (
	"context"
	"github.com/chromedp/cdproto/cdp"
	"github.com/chromedp/cdproto/dom"
	"github.com/chromedp/cdproto/runtime"
	"github.com/chromedp/chromedp"
	"log"
)

func main() {
	options := append(chromedp.DefaultExecAllocatorOptions[:], chromedp.Flag("headless", false), chromedp.WindowSize(1280, 1024))
	ctx, cancel := chromedp.NewExecAllocator(context.Background(), options...)
	ctx, cancel = chromedp.NewContext(ctx, chromedp.WithDebugf(log.Printf))
	defer cancel()
	var res interface{}
	log.Println(chromedp.Run(ctx,
		chromedp.Navigate("https://github.com/"),
		chromedp.ActionFunc(func(ctx context.Context) error {
			err := chromedp.CallFunctionOn(`(data)=>{console.log(data); return null}`, &res, func(params *runtime.CallFunctionOnParams) *runtime.CallFunctionOnParams {
				c := chromedp.FromContext(ctx)
				ctx = cdp.WithExecutor(ctx, c.Target)
				nodes := make([]*cdp.Node, 0)
				if err := chromedp.Nodes("html", &nodes).Do(ctx); err != nil {
					return params.WithSilent(false)
				}
				r, err := dom.ResolveNode().WithNodeID(nodes[0].NodeID).Do(ctx)
				if err != nil {
					return params.WithSilent(false)
				}
				return params.WithReturnByValue(true).WithObjectID(r.ObjectID)
			}, map[string]interface{}{"a": 1, "b": "b", "c": true}).Do(ctx)
			if err != nil {
				panic(err) // unexpected end of JSON input
			}
			return nil
		}),
	))
}

```

#### What did you expect to see?
Since the return value of js is null, the result of res of the corresponding go should be nil instead of returning an error.
#### What did you see instead?
It will panic whith unexpected end of JSON input.
[...]

PR #1260, opened 2023-02-14
compatible object null resolves to go type

xueyc1f:
I tracked the CDP log and found that when js expression return is null, then `runtime.RemoteObject.Type` is object and `runtime.RemoteObject.Value` is nil, so I added a judgment and returned in advance
#1259

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -79,16 +79,11 @@ func parseRemoteObject(v *runtime.RemoteObject, res interface{}) (err error) {
 	value := v.Value
 	if v.Type == "undefined" || value == nil {
 		rv := reflect.ValueOf(res)
-	checkKind:
 		if rv.Kind() == reflect.Pointer {
 			rv = rv.Elem()
 			switch rv.Kind() {
-			case reflect.Map, reflect.Pointer, reflect.Slice, reflect.Func, reflect.Chan:
-				// map, ptr, slice, func, chan type can be handled correctly by json package
-			case reflect.Interface:
-				// if `res` is interface, get its actual type and check again
-				rv = rv.Elem()
-				goto checkKind
+			case reflect.Map, reflect.Pointer, reflect.Slice, reflect.Func, reflect.Chan, reflect.Interface:
+				// map, ptr, slice, func, chan, interface type can be handled correctly by json package
 			default:
 				// return [ErrJSUndefined] or [ErrJSNull] respectively.
 				// the caller maybe need to redefined `res` type
--- a/eval.go
+++ b/eval.go
@@ -29,9 +29,9 @@ type EvaluateAction Action
 // For all other cases, the result of the script will be returned "by value" (i.e.,
 // JSON-encoded), and subsequently an attempt will be made to json.Unmarshal
 // the script result to res. When the script result is "undefined" or "null",
-// and the value that res points to is not a pointer, or it points to a
-// map/ptr/slice/interface/func/chan, it will handle by json.Unmarshal.
-// for other type it returns [ErrJSUndefined] of [ErrJSNull] respectively.
+// and the value that res points to can not be nil (only the value of a chan,
+// func, interface, map, pointer, or slice can be nil), it returns [ErrJSUndefined]
+// or [ErrJSNull] respectively.
 func Evaluate(expression string, res interface{}, opts ...EvaluateOption) EvaluateAction {
 	return ActionFunc(func(ctx context.Context) error {
 		// set up parameters
@@ -80,18 +80,21 @@ func parseRemoteObject(v *runtime.RemoteObject, res interface{}) (err error) {
 		rv := reflect.ValueOf(res)
 		if rv.Kind() == reflect.Pointer {
 			switch rv.Elem().Kind() {
-			case reflect.Map, reflect.Pointer, reflect.Slice, reflect.Func, reflect.Chan, reflect.Interface:
-				// map, ptr, slice, func, chan, interface type can be handled correctly by json package
+			// Common kinds that can be nil.
+			case reflect.Pointer, reflect.Map, reflect.Slice:
+			// It's weird that res is a pointer to the following kinds,
+			// but they can be nil too.
+			case reflect.Chan, reflect.Func, reflect.Interface:
 			default:
+				// When the value that `res` points to can not be set to nil,
 				// return [ErrJSUndefined] or [ErrJSNull] respectively.
-				// the caller maybe need to redefined `res` type
 				if v.Type == "undefined" {
 					return ErrJSUndefined
 				}
 				return ErrJSNull
 			}
 		}
-		// `res` is not a pointer, change the value to the json literal null, returns error by json.Unmarshal
+		// Change the value to the json literal null to make json.Unmarshal happy.
 		value = []byte("null")
 	}
 
--- a/eval.go
+++ b/eval.go
@@ -75,28 +75,28 @@ func parseRemoteObject(v *runtime.RemoteObject, res interface{}) (err error) {
 		return
 	}
 
-	if v.Type == "undefined" || v.Value == nil {
-		// `res` should be a pointer. When the value that `res` points to is
-		// not a pointer, it can not be nil. In this case,
-		// return [ErrJSUndefined] or [ErrJSNull] respectively.
-		rv := reflect.Indirect(reflect.ValueOf(res))
-		switch rv.Kind() {
-		// json.Unmarshal supported in this case types
-		case reflect.Array, reflect.Chan, reflect.Func, reflect.Interface, reflect.Map,
-			reflect.Pointer, reflect.Slice, reflect.Struct:
-			// Otherwise, change the value to the json literal null.
-			v.Value = []byte("null")
-		default:
-			// other type zero value isn't nil, return error
-			if v.Type == "undefined" {
-				return ErrJSUndefined
-			}
-			return ErrJSNull
+	value := v.Value
+	if value == nil {
+		// if js expression is undefined, it cannot resolve to any go type, so return ErrJSUndefined directly
+		if v.Type == "undefined" {
+			return ErrJSUndefined
 		}
+		rv := reflect.ValueOf(res)
+		if rv.Kind() == reflect.Pointer {
+			switch rv.Elem().Kind() {
+			case reflect.Interface, reflect.Map, reflect.Pointer, reflect.Slice:
+				// only interface, map, ptr, slice type can resolve js null
+				value = []byte("null")
+			default:
+				// other types return ErrJSNull, the caller maybe need to redefined `res` type
+				return ErrJSNull
+			}
+		}
+		// `res` is not a pointer, change the value to the json literal null, errored by json.Unmarshal
+		value = []byte("null")
 	}
 
-	err = json.Unmarshal(v.Value, res)
-	return
+	return json.Unmarshal(value, res)
 }
 
 // EvaluateAsDevTools is an action that evaluates a JavaScript expression as
--- a/eval.go
+++ b/eval.go
@@ -90,6 +90,13 @@ func parseRemoteObject(v *runtime.RemoteObject, res interface{}) (undefined bool
 		return
 	}
 
+	if v.Type == runtime.TypeObject && v.Value == nil {
+		// when exception `return null`, v.Type must be runtime.TypeObject and
+		// v.Value is nil, res and error should be nil
+		res = nil
+		return false, nil
+	}
+
 	// unmarshal
 	err = json.Unmarshal(v.Value, res)
 	return
--- a/eval.go
+++ b/eval.go
@@ -76,7 +76,7 @@ func parseRemoteObject(v *runtime.RemoteObject, res interface{}) (err error) {
 	}
 
 	value := v.Value
-	if v.Type == "undefined" || value == nil {
+	if value == nil {
 		rv := reflect.ValueOf(res)
 		if rv.Kind() == reflect.Pointer {
 			switch rv.Elem().Kind() {
--- a/eval.go
+++ b/eval.go
@@ -56,8 +56,7 @@ func Evaluate(expression string, res interface{}, opts ...EvaluateOption) Evalua
 			return exp
 		}
 
-		err = parseRemoteObject(v, res)
-		return err
+		return parseRemoteObject(v, res)
 	})
 }
 
@@ -80,8 +79,7 @@ func parseRemoteObject(v *runtime.RemoteObject, res interface{}) (err error) {
 	if v.Type == "undefined" || value == nil {
 		rv := reflect.ValueOf(res)
 		if rv.Kind() == reflect.Pointer {
-			rv = rv.Elem()
-			switch rv.Kind() {
+			switch rv.Elem().Kind() {
 			case reflect.Map, reflect.Pointer, reflect.Slice, reflect.Func, reflect.Chan, reflect.Interface:
 				// map, ptr, slice, func, chan, interface type can be handled correctly by json package
 			default:
--- a/eval.go
+++ b/eval.go
@@ -3,7 +3,7 @@ package chromedp
 import (
 	"context"
 	"encoding/json"
-	"fmt"
+	"reflect"
 
 	"github.com/chromedp/cdproto/runtime"
 )
@@ -22,18 +22,15 @@ type EvaluateAction Action
 //
 // When res is a **runtime.RemoteObject, res will be set to the low-level
 // protocol type, and no attempt will be made to convert the result.
-// Original objects are maintained in memory until the page navigated or closed,
-// unless they are either explicitly released or are released along with the
-// other objects in their object group. runtime.ReleaseObject or
-// runtime.ReleaseObjectGroup can be used to ask the browser to release
-// original objects.
+// The original objects could be maintained in memory until the page is
+// navigated or closed. `runtime.ReleaseObject` or `runtime.ReleaseObjectGroup`
+// can be used to ask the browser to release the original objects.
 //
 // For all other cases, the result of the script will be returned "by value" (i.e.,
 // JSON-encoded), and subsequently an attempt will be made to json.Unmarshal
-// the script result to res. It returns an error if the script result is
-// "undefined" in this case.
-//
-// Note: any exception encountered will be returned as an error.
+// the script result to res. When the script result is "undefined" or "null",
+// and the value that res points to is not a pointer, it returns [ErrJSUndefined]
+// of [ErrJSNull] respectively.
 func Evaluate(expression string, res interface{}, opts ...EvaluateOption) EvaluateAction {
 	return ActionFunc(func(ctx context.Context) error {
 		// set up parameters
@@ -58,16 +55,12 @@ func Evaluate(expression string, res interface{}, opts ...EvaluateOption) Evalua
 			return exp
 		}
 
-		_, err = parseRemoteObject(v, res)
+		err = parseRemoteObject(v, res)
 		return err
 	})
 }
 
-func parseRemoteObject(v *runtime.RemoteObject, res interface{}) (undefined bool, err error) {
-	// undefined indicates that the result is a javascript "undefined" value.
-	// Poll needs this value to decide whether it's a timeout.
-	undefined = v.Type == "undefined"
-
+func parseRemoteObject(v *runtime.RemoteObject, res interface{}) (err error) {
 	if res == nil {
 		return
 	}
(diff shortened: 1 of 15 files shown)


ZekeLu:
@xueyc1f Thank you for the contribution!

I just sent #1261. Which I think handle it more correctly. Can you review it and give it a try?

xueyc1f:
@ZekeLu I don't think this problem can be solved perfectly at present. If res type is int or string and other basic types, it is correct at present, but if res type is map and other types, it will still return an error `encountered a null value`

The code of:
```go
var res = map[string]interface{}{}
err := chromedp.CallFunctionOn(`(data)=>{console.log(data); return null}`, &res, optFunc, map[string]interface{}{"a": 1, "b": "b", "c": true})
// err is encountered a null value
// if var res = &map[string]interface{}{}, CallFunctionOn is correct.
err = json.Unmarshal([]byte("null"), &res)
// err is null, so CallFunctionOn hould be handled like json.Unmarshal
```

ZekeLu:
You are correct. A chan, func, interface, map, pointer, and slice value could be nil. I think we should handle map and slice too. Would you like to send a fix?
Thank you!

xueyc1f:
Sure, I can submit fix later

xueyc1f:
Hi @ZekeLu , I tested that json.Unmarshal supports such parsing, so I think parseRemoteObject should also provide similar compatibility. If v.Value is not null, the error is returned by json.Unmarshal
```go
var i chan struct{}
err := json.Unmarshal([]byte(`null`), &i) // err is nil

err := json.Unmarshal([]byte(`{}`), &i) // err is json: cannot unmarshal object into Go value of type chan struct {}
```

eval.go
```diff
@@ -76,18 +76,23 @@ func parseRemoteObject(v *runtime.RemoteObject, res interface{}) (err error) {
...
 	if v.Type == "undefined" || v.Value == nil {
-		rv := reflect.ValueOf(res)
 		// `res` should be a pointer. When the value that `res` points to is
 		// not a pointer, it can not be nil. In this case,
 		// return [ErrJSUndefined] or [ErrJSNull] respectively.
-		if rv.Kind() == reflect.Pointer && rv.Elem().Kind() != reflect.Pointer {
+		rv := reflect.Indirect(reflect.ValueOf(res))
+		switch rv.Kind() {
+		// json.Unmarshal supported in this case types
+		case reflect.Array, reflect.Chan, reflect.Func, reflect.Interface, reflect.Map,
```
ZekeLu, line 85: I think the zero value of `reflect.Array` and `reflect.Struct` is not nil.

eval_test.go
```diff
@@ -204,3 +205,59 @@ func TestEvaluateNil(t *testing.T) {
 		})
 	}
 }
+
+func TestEvaluateNull(t *testing.T) {
```
ZekeLu, line 209: It's okay to add a test here. But I want to list the common use cases in `ExampleEvaluate` for documentation purpose.

image

ZekeLu (changes requested):
```go
var i int
err := Run(ctx, Evaluate("null", i))
```

After this change, the `err` in the above code will be `ErrJSNull`, but I think it's better to report it as `json: Unmarshal(non-pointer int)`.

If we just need to handle the `undefined` and `null` cases, the code could simply be:

```go
value := v.Value
if value == nil {
	// Change the value to the json literal null to make json.Unmarshal happy.
	value = []byte("null")
}

return json.Unmarshal(value, res)
```

But for the following code,

```go
var i int
err := Run(ctx, Evaluate("undefined", &i))
```

We want the error to be `ErrJSUndefined`. See this test:

https://github.com/chromedp/chromedp/blob/2c1b71999a2f16e3e121672c3a0727cef367ca18/query_test.go#L595-L608

So I would suggest to change it like this:

```go
value := v.Value

if value == nil {
	rv := reflect.ValueOf(res)
	if rv.Kind() == reflect.Pointer {
		switch rv.Elem().Kind() {
		// Common kinds that can be nil.
		case reflect.Pointer, reflect.Map, reflect.Slice:
		// It's weird that res is a pointer to the following kinds,
		// but they can be nil too.
		case reflect.Chan, reflect.Func, reflect.Interface:
		default:
			// When the value that `res` points to can not be set to nil,
			// return [ErrJSUndefined] or [ErrJSNull] respectively.
			if v.Type == "undefined" {
				return ErrJSUndefined
			}
			return ErrJSNull
		}
	}

	// Change the value to the json literal null to make json.Unmarshal happy.
	value = []byte("null")
}

return json.Unmarshal(value, res)
```

xueyc1f:
@ZekeLu Based on your suggestions and the defect of go json: Unmarshal null [golang/go/issues/33835], I summarized the following processing of parseRemoteObject to go type:
- If js expression is undefined, it cannot resolve to any go type, so return ErrJSUndefined directly
- If js expression is null, it can only resolve to interface, map, ptr, slice type. If other types such as array, string and other zero values cannot be nil, and chan, func cannot be serialized to json, return ErrJSNull 
Do you have any other supplements or suggestions?

ZekeLu:
I would consider this problem from the perspective of the client of `chromedp`.

1. If the user code is like this:

```go
var i int
err := chromedp.Run(ctx, chromedp.Evaluate("any js expression", i))
```

It's a mistake and we should let the user know. The `json` package can report such kind of error. (`json: Unmarshal(non-pointer int)`. That's perfect. So let the `json` package do the job.

2. If the user code is like this:

```go
var i int
err := chromedp.Run(ctx, chromedp.Evaluate("js expression that will never evaluate to undefined/null", &i))
```

I would assume that the user don't expect undefined/null at all. If it happens, we should report `ErrJSUndefined` or `ErrJSNull` to the user. If the user thinks that it's fine to receive undefined/null, then the user should write the code like this:

```go
var i *int
err := chromedp.Run(ctx, chromedp.Evaluate("js expression that evaluates to integer/undefined/null", &i))
```

3. If it doesn't work at all, the user can get the remote object and handle it himself/herself:

```go
var res *runtime.RemoteObject
err := chromedp.Run(ctx, chromedp.Evaluate("any js expression", &res))
```

xueyc1f:
I sent a new submission. I think it seems to meet your above settings. If have any questions, please let me know

eval.go
```diff
@@ -75,23 +75,28 @@ func parseRemoteObject(v *runtime.RemoteObject, res interface{}) (err error) {
...
 		rv := reflect.ValueOf(res)
-		// `res` should be a pointer. When the value that `res` points to is
-		// not a pointer, it can not be nil. In this case,
-		// return [ErrJSUndefined] or [ErrJSNull] respectively.
-		if rv.Kind() == reflect.Pointer && rv.Elem().Kind() != reflect.Pointer {
-			if v.Type == "undefined" {
-				return ErrJSUndefined
+		if rv.Kind() == reflect.Pointer {
+			switch rv.Elem().Kind() {
+			case reflect.Interface, reflect.Map, reflect.Pointer, reflect.Slice:
```
ZekeLu, line 87:
```go
var f func()
err := Run(ctx, Evaluate("null", &f))
```

It seems that returning ErrJSNull in this case does not help. I think we can leave it to the `json` package.

ZekeLu (changes requested):
Please don't forget to update the doc comment for `Evaluate`:

https://github.com/chromedp/chromedp/blob/2c1b71999a2f16e3e121672c3a0727cef367ca18/eval.go#L15-L34

eval.go
```diff
@@ -75,23 +75,28 @@ func parseRemoteObject(v *runtime.RemoteObject, res interface{}) (err error) {
 		return
 	}
 
-	if v.Type == "undefined" || v.Value == nil {
+	value := v.Value
+	if value == nil {
+		// if js expression is undefined, it cannot resolve to any go type, so return ErrJSUndefined directly
+		if v.Type == "undefined" {
+			return ErrJSUndefined
+		}
```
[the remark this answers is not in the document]
xueyc1f: I don't think we should be able to distinguish what js code is undefined,undefined does not resolve to go type, so we should return the ErrJSUndefined error, and the caller can choose to ignore the error or check his js code

eval.go
xueyc1f: var f func()
err := json.Unmarshal([]byte("null"), &f) // err is nil
chan, func cannot be unserialized by json, so we should return ErrJSNull

eval.go
```diff
@@ -75,23 +75,28 @@ func parseRemoteObject(v *runtime.RemoteObject, res interface{}) (err error) {
...
-		// not a pointer, it can not be nil. In this case,
-		// return [ErrJSUndefined] or [ErrJSNull] respectively.
-		if rv.Kind() == reflect.Pointer && rv.Elem().Kind() != reflect.Pointer {
-			if v.Type == "undefined" {
-				return ErrJSUndefined
+		if rv.Kind() == reflect.Pointer {
+			switch rv.Elem().Kind() {
+			case reflect.Interface, reflect.Map, reflect.Pointer, reflect.Slice:
+				// only interface, map, ptr, slice type can resolve js null
+				value = []byte("null")
```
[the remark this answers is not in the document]
xueyc1f: yes

eval.go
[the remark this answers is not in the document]
xueyc1f: `var m = new Map(); m.get('key')`  evaluate to undefined, so it return `ErrJSUndefined` is right

image

It seems to be exactly what you said：
> I would assume that the user don't expect undefined/null at all. If it happens, we should report ErrJSUndefined or ErrJSNull to the user. If the user thinks that it's fine to receive undefined/null, then the user should write the code like this:
> 
> ```go
> var i *int
> err := chromedp.Run(ctx, chromedp.Evaluate("js expression that evaluates to integer/undefined/null", &i))
> ```
>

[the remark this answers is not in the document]
xueyc1f: > And if we want this, I think we should move this code block above value := v.Value.
I agree with you, but I'm not sure whether there is `v.Type=="undefined" && v.Value != nil` situation.
I assume that it exists then handle by json package.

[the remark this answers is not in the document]
ZekeLu: > It seems to be exactly what you said：
>> ... If the user thinks that it's fine to receive undefined/null, then the user should write the code like this:

It seems that you don't get my point. Imagine this scenario:

```go
var i *int
err := Run(ctx,
	Evaluate("some expression evaluated to undefined or null or an integer", &i),
	// other actions below
)
```

By saying "it's fine to receive undefined/null", I meant that even if the js result is `undefined`, __// other actions below__ should be executed.

This is not true if it returns `ErrJSUndefined` whenever the js result is `undefined`. And if forces me to change the code to:

```go
var i *int
err := Run(ctx,
	Evaluate("some expression evaluated to undefined or null or an integer", &i),
)

if err == nil || err == ErrJSUndefined {
	err = Run(ctx,
		// other actions below
	)
}
```

> but I'm not sure whether there is `v.Type=="undefined" && v.Value != nil` situation. I assume that it exists then handle by json package.

I think it's safe to say that `v.Value` is `nil` when `v.Type == "undefined"`.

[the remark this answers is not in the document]
ZekeLu: > If js expression is undefined, it cannot resolve to any go type, so return ErrJSUndefined directly

I think you meant this:

```go
var i int
err := json.Unmarshal([]byte("undefined"), &i)
// err: invalid character 'i' in literal null (expecting 'u'), i: 0
```

That's because "undefined" is not a valid json (see https://www.json.org/json-en.html). But that does not mean that `undefined` is not supported by the json package.

For example, I would say that `b` is `undefined` in the json `{"a":"valueA"}` (see the js code below).

```js
var obj = JSON.parse('{"a":"valueA"}');
console.log(typeof obj.b);
// undefined
```

And the json package can unmarshal `undefined` value:

```go
type UndefinedAsZeroValue struct {
	A string `json:"a"`
	B int    `json:"b"`
}
var obj1 UndefinedAsZeroValue
err := json.Unmarshal([]byte(`{"a":"valueA"}`), &obj1)
fmt.Printf("err: %v, obj1: %#v\n", err, obj1)
// err: <nil>, obj1: chromedp.UndefinedAsZeroValue{A:"valueA", B:0}

type UndefinedAsNil struct {
	A string `json:"a"`
	B *int   `json:"b"`
}
var obj2 UndefinedAsNil
err = json.Unmarshal([]byte(`{"a":"valueA"}`), &obj2)
fmt.Printf("err: %v, obj2: %#v\n", err, obj2)
// err: <nil>, obj2: chromedp.UndefinedAsNil{A:"valueA", B:(*int)(nil)}
```

eval.go
ZekeLu: See the code below:

```go
var f func()
err := json.Unmarshal([]byte("null"), &f)
fmt.Println(err)
// <nil>
err = Run(ctx, Evaluate("null", &f))
fmt.Println(err)
// encountered a null value

err = json.Unmarshal([]byte("{}"), &f)
fmt.Println(err)
// json: cannot unmarshal object into Go value of type func()

err = Run(ctx, Evaluate("({})", &f))
fmt.Println(err)
// json: cannot unmarshal object into Go value of type func()
```

`null` is allowed by the json package. I don't understand why we should do something different. What is the advantage of returning `ErrJSNull` in this case?

xueyc1f: for array and struct type, should we also not return errors?
```go
var arr [3]int
err := json.Unmarshal([]byte(`null`), &arr) // err is nil

var s struct{}
err := json.Unmarshal([]byte(`null`), &s) // err is nil
```

ZekeLu: The zero value of array and struct are not nil, so we will treat them like int.

eval.go
```diff
@@ -76,23 +77,29 @@ func parseRemoteObject(v *runtime.RemoteObject, res interface{}) (err error) {
...
-				value = []byte("null")
+			rv = rv.Elem()
+			switch rv.Kind() {
+			case reflect.Map, reflect.Pointer, reflect.Slice,
+				reflect.Func, reflect.Chan, reflect.Struct, reflect.Array:
+				// map, ptr, slice, func, chan, struct, array type can be handled correctly by json package
+			case reflect.Interface:
+				// if `res` is interface, get its actual type and check again
+				rv = rv.Elem()
+				goto checkKind
```
ZekeLu, line 91: Can we have a test case for this code path?

xueyc1f: yes, test at eval_test.goL244

ZekeLu: That test does not show why we need

```go
rv = rv.Elem()
goto checkKind
```

xueyc1f: query_test.go `TestValueUndefined` will be executed to this case

xueyc1f: func `Value` calls the func `JavascriptAttribute` the parameter res is converted from type * string to interface {}, so we need the above code

ZekeLu: I see. That's a bug in `callFunctionOnNode`. I have just sent #1264 to fix it. Thank you!

I think we don't need to treat `reflect.Interface` differently now. What do you think?

xueyc1f: Yes, not now

example_test.go
```diff
@@ -628,6 +628,17 @@ func ExampleEvaluate() {
 		}
 	}
 
+	// Accept undefined/nil result by ptr:
```
ZekeLu, line 631: Let's change the example to:

```go
func ExampleEvaluate() {
	defer cancel()

	// Ignore the result:
	{
		if err := chromedp.Run(ctx,
			chromedp.Evaluate(`window.scrollTo(0, 100)`, nil),
		); err != nil {
			log.Fatal(err)
		}
	}

	// Receive a primary value:
	{
		var sum int
		if err := chromedp.Run(ctx,
			chromedp.Evaluate(`1+2`, &sum),
		); err != nil {
			log.Fatal(err)
		}
		fmt.Println(sum)
	}

	// ErrJSUndefined:
	{
		var val int
		if err := chromedp.Run(ctx,
			chromedp.Evaluate(`undefined`, &val),
		); err != nil {
			fmt.Println(err)
		}
	}

	// ErrJSNull:
	{
		var val int
		if err := chromedp.Run(ctx,
			chromedp.Evaluate(`null`, &val),
		); err != nil {
			fmt.Println(err)
		}
	}

	// Accept undefined/null result:
	{
		var val *int
		if err := chromedp.Run(ctx,
			chromedp.Evaluate(`undefined`, &val),
		); err != nil {
			log.Fatal(err)
		}
		fmt.Println(val)
	}

	// Receive an array value:
	{
		var val []int
		if err := chromedp.Run(ctx,
			chromedp.Evaluate(`[1,2]`, &val),
		); err != nil {
			log.Fatal(err)
		}
		fmt.Println(val)
	}

	// Map and Slice accept undefined/null:
	{
		var val []int
		if err := chromedp.Run(ctx,
			chromedp.Evaluate(`null`, &val),
		); err != nil {
			log.Fatal(err)
		}
		fmt.Println("slice is nil:", val == nil)
	}

	// Receive the raw bytes:
	{
		var buf []byte
		if err := chromedp.Run(ctx,
			chromedp.Evaluate(`alert`, &buf),
		); err != nil {
			log.Fatal(err)
		}
		fmt.Printf("%s\n", buf)
	}

	// Receive the RemoteObject:
	{
		var res *runtime.RemoteObject
		if err := chromedp.Run(ctx,
			chromedp.Evaluate(`alert`, &res),
		); err != nil {
			log.Fatal(err)
		}
		if res.ObjectID != "" {
			fmt.Println("objectId is present")
		}
	}

	// Output:
	// 3
	// encountered an undefined value
	// encountered a null value
	// <nil>
	// [1 2]
	// slice is nil: true
	// {}
	// objectId is present
}
```

eval.go
```diff
@@ -75,23 +75,27 @@ func parseRemoteObject(v *runtime.RemoteObject, res interface{}) (err error) {
 		return
 	}
 
-	if v.Type == "undefined" || v.Value == nil {
+	value := v.Value
```
ZekeLu, line 78: Let's modify this part to:

```go
	value := v.Value

	if value == nil {
		rv := reflect.ValueOf(res)
		if rv.Kind() == reflect.Pointer {
			switch rv.Elem().Kind() {
			// Common kinds that can be nil.
			case reflect.Pointer, reflect.Map, reflect.Slice:
			// It's weird that res is a pointer to the following kinds,
			// but they can be nil too.
			case reflect.Chan, reflect.Func, reflect.Interface:
			default:
				// When the value that `res` points to can not be set to nil,
				// return [ErrJSUndefined] or [ErrJSNull] respectively.
				if v.Type == "undefined" {
					return ErrJSUndefined
				}
				return ErrJSNull
			}
		}

		// Change the value to the json literal null to make json.Unmarshal happy.
		value = []byte("null")
	}

	return json.Unmarshal(value, res)
```

eval.go
```diff
@@ -29,8 +29,9 @@ type EvaluateAction Action
 // For all other cases, the result of the script will be returned "by value" (i.e.,
 // JSON-encoded), and subsequently an attempt will be made to json.Unmarshal
 // the script result to res. When the script result is "undefined" or "null",
-// and the value that res points to is not a pointer, it returns [ErrJSUndefined]
-// of [ErrJSNull] respectively.
+// and the value that res points to is not a pointer, or it points to a
+// map/ptr/slice/interface/func/chan, it will handle by json.Unmarshal.
+// for other type it returns [ErrJSUndefined] of [ErrJSNull] respectively.
```
ZekeLu, line 34: Let's change the comment block to:

```go
// For all other cases, the result of the script will be returned "by value" (i.e.,
// JSON-encoded), and subsequently an attempt will be made to json.Unmarshal
// the script result to res. When the script result is "undefined" or "null",
// and the value that res points to can not be nil (only the value of a chan,
// func, interface, map, pointer, or slice can be nil), it returns [ErrJSUndefined]
// or [ErrJSNull] respectively.
```

ZekeLu (changes requested):
Please take a look at my suggested changes. Thanks!

xueyc1f:
> 

Thanks for your suggestion, it's better

eval.go
```diff
@@ -75,23 +75,30 @@ func parseRemoteObject(v *runtime.RemoteObject, res interface{}) (err error) {
 		return
 	}
 
-	if v.Type == "undefined" || v.Value == nil {
+	value := v.Value
+	if v.Type == "undefined" || value == nil {
```
ZekeLu, line 79: Please change to:

```
if value == nil {
```

`value` should be nil when `v.Type == "undefined"`. If not, we should not try to hide that.

ZekeLu (changes requested):

eval.go
xueyc1f: yes

ZekeLu (approved):

PR #1261, opened 2023-02-14
make parseRemoteObject handle undefined/null reasonably

### LINKED ISSUE #1259: CallFunctionOn does not properly handle the null returned by js

ZekeLu:
When the value that `res` points to is a pointer, it should accept `undefined` or  `null`.

```go
	// ErrJSUndefined:
	{
		var val int
		if err := chromedp.Run(ctx,
			chromedp.Evaluate(`undefined`, &val),
		); err != nil {
			fmt.Println(err)
		}
	}

	// ErrJSNull:
	{
		var val int
		if err := chromedp.Run(ctx,
			chromedp.Evaluate(`null`, &val),
		); err != nil {
			fmt.Println(err)
		}
	}

	// Accept undefined/nil result:
	{
		var val *int
		if err := chromedp.Run(ctx,
			chromedp.Evaluate(`undefined`, &val),
		); err != nil {
			log.Fatal(err)
		}
		fmt.Println(val)
	}
```

Fixes #1259.

### DIFF
--- a/call.go
+++ b/call.go
@@ -28,7 +28,7 @@ func CallFunctionOn(functionDeclaration string, res interface{}, opt CallOption,
 	})
 }
 
-func callFunctionOn(ctx context.Context, functionDeclaration string, res interface{}, opt CallOption, args ...interface{}) (bool, error) {
+func callFunctionOn(ctx context.Context, functionDeclaration string, res interface{}, opt CallOption, args ...interface{}) (*runtime.RemoteObject, error) {
 	// set up parameters
 	p := runtime.CallFunctionOn(functionDeclaration).
 		WithSilent(true)
@@ -51,7 +51,7 @@ func callFunctionOn(ctx context.Context, functionDeclaration string, res interfa
 			ea.append(arg)
 		}
 		if ea.err != nil {
-			return false, ea.err
+			return nil, ea.err
 		}
 		p = p.WithArguments(ea.args)
 	}
@@ -59,13 +59,13 @@ func callFunctionOn(ctx context.Context, functionDeclaration string, res interfa
 	// call
 	v, exp, err := p.Do(ctx)
 	if err != nil {
-		return false, err
+		return nil, err
 	}
 	if exp != nil {
-		return false, exp
+		return nil, exp
 	}
 
-	return parseRemoteObject(v, res)
+	return v, parseRemoteObject(v, res)
 }
 
 // CallOption is a function to modify the runtime.CallFunctionOnParams
--- a/chromedp_test.go
+++ b/chromedp_test.go
@@ -1075,7 +1075,7 @@ func TestWebGL(t *testing.T) {
 		Screenshot(`#c`, &buf, ByQuery),
 	); err != nil {
 		if errors.Is(err, ErrPollingTimeout) {
-			t.Fatal("The cube it not rendered in 2s.")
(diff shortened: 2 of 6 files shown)


kenshaw:
@ZekeLu I think this is reasonable; I've reviewed the logic already, and rebased it. We can tag a release if we find this to be stable / working. I'd suggest trying it with some more complicated examples. At the end of the day, the person writing a script for `Eval` would need to know what their script does, and how `chromedp` in turn marshals results.

ZekeLu:
Thank you Ken!

> I'd suggest trying it with some more complicated examples.

Unfortunately, my projects that depend on `chromedp` don't have a complicated use case of the related functions. So I can't test it thoroughly.

PR #1295, opened 2023-04-11
Fixed nil pointer exception when cancelling a context with chromedp.Cancel()

### LINKED ISSUE #1294: Nil pointer exception when cancelling a context with chromedp.Cancel()

```console
github.com/chromedp/chromedp v0.9.1
$ google-chrome --version
Version 112.0.5615.50 (Offizieller Build) (64-Bit)
$ go version
go version go1.20.1 windows/amd64
```

```go
opts := append(
	chromedp.DefaultExecAllocatorOptions[:],
	chromedp.Flag("ignore-certificate-errors", true),
)

if debug {
	opts = append(opts, chromedp.Flag("headless", false))
}

globalContext, _ = chromedp.NewExecAllocator(context.Background(), opts...)

if remoteRendering {
	globalContext, _ = chromedp.NewRemoteAllocator(
		globalContext,
		config.Get().RemoteRenderingURL,
	)
}

...

ctx, _ = chromedp.NewContext(
		globalContext,
		chromedp.WithDebugf(func(s string, i ...interface{}) {}),
		chromedp.WithLogf(func(s string, i ...interface{}) {}),
		chromedp.WithErrorf(func(s string, i ...interface{}) {}),
	) // Disable logging

err = chromedp.Run(ctx, ...)
...

err := chromedp.Cancel(globalContext)
if err != nil {
...
}
```

The context should have been canceled or an error should have been returned.

A nil pointer exception:
```
runtime error: invalid memory address or nil pointer dereference
.../go/pkg/mod/github.com/chromedp/chromedp@v0.9.1/chromedp.go:257
```

I had the context printed out, here is the result before c.cancel() is called in chromedp.go:257
```
&{Allocator:0xc0004266c0 Browser:<nil> Target:<nil> targetID: createBrowserContextParams:<nil> browserContextOwner:false BrowserContextID: browserListeners:[] targetListeners:[] browserOpts:[] cancel:<nil> first:false closedTarget:{noCopy:{} state:{_:{} _:{} v:0} sema:0} allocated:<nil> cancelErr:<nil>}
```

Valentin-Kaiser:
Fixes #1294

### DIFF
--- a/chromedp.go
+++ b/chromedp.go
@@ -244,6 +244,8 @@ func FromContext(ctx context.Context) *Context {
 // underlying errors happening during cancellation.
 func Cancel(ctx context.Context) error {
 	c := FromContext(ctx)
+	// c.cancel is nil when Cancel is wrongly called with a context returned
+	// by chromedp.NewExecAllocator or chromedp.NewRemoteAllocator.
 	if c == nil || c.cancel == nil {
 		return ErrInvalidContext
 	}
--- a/chromedp.go
+++ b/chromedp.go
@@ -244,7 +244,7 @@ func FromContext(ctx context.Context) *Context {
 // underlying errors happening during cancellation.
 func Cancel(ctx context.Context) error {
 	c := FromContext(ctx)
-	if c == nil {
+	if c == nil || c.cancel == nil {
 		return ErrInvalidContext
 	}
 	graceful := c.first && c.Browser != nil
--- a/chromedp_test.go
+++ b/chromedp_test.go
@@ -261,6 +261,12 @@ func TestCancelError(t *testing.T) {
 		t.Fatalf("expected a nil error, got %v", err)
 	}
 
+	allocCtx, allocCancel := NewExecAllocator(context.Background(), allocOpts...)
+	defer allocCancel()
+	if err := Cancel(allocCtx); err != ErrInvalidContext {
+		t.Fatalf("want error %q, got %q", ErrInvalidContext, err)
+	}
+
 	/*
 		// NOTE: the following test no longer is applicable, as a slight change
 		// to chromium's 89 API deprecated the boolean return value
--- a/chromedp_test.go
+++ b/chromedp_test.go
@@ -261,8 +261,6 @@ func TestCancelError(t *testing.T) {
 		t.Fatalf("expected a nil error, got %v", err)
 	}
 
-	allocCtx, allocCancel := NewExecAllocator(context.Background(), allocOpts...)
-	defer allocCancel()
 	if err := Cancel(allocCtx); err != ErrInvalidContext {
 		t.Fatalf("want error %q, got %q", ErrInvalidContext, err)
 	}
--- a/allocate.go
+++ b/allocate.go
@@ -87,7 +87,7 @@ var DefaultExecAllocatorOptions = [...]ExecAllocatorOption{
 // for use with NewContext.
 func NewExecAllocator(parent context.Context, opts ...ExecAllocatorOption) (context.Context, context.CancelFunc) {
 	ctx, cancel := context.WithCancel(parent)
-	c := &Context{Allocator: setupExecAllocator(opts...), cancel: cancel}
+	c := &Context{Allocator: setupExecAllocator(opts...)}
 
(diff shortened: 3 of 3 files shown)


chromedp.go
```diff
@@ -244,7 +244,7 @@ func FromContext(ctx context.Context) *Context {
 // underlying errors happening during cancellation.
 func Cancel(ctx context.Context) error {
 	c := FromContext(ctx)
-	if c == nil {
+	if c == nil || c.cancel == nil {
```
ZekeLu, line 247: I think we only need to add a check before `c.cancel()` on line 257.

```diff
    } else {
+       // c.cancel is nil when Cancel is wrongly called with a context returned
+       // from chromedp.NewRemoteAllocator or chromedp.NewRemoteAllocator.
+       if c.cancel == nil {
+           return ErrInvalidContext
+       }
        c.cancel()
        c.closedTarget.Wait()
    }
```

ZekeLu (changes requested):
Issue #1294 is a user error, we can try our best to let the user know what's wrong. But we should not complicate the code solely for this purpose. So I think we just need to add a check before `c.cancel()`.

chromedp.go
Valentin-Kaiser: But `c.cancel()` is called again in line 275, so the error could occur again there.

Furthermore, this error no longer occurs if cancel is passed to the context in chromedp.NewRemoteAllocator or chromedp.NewExecAllocator.

And the correct comment would be:
```
// c.cancel is nil when Cancel is wrongly called with a context returned 
// from chromedp.NewRemoteAllocator or chromedp.NewExecAllocator
```

ZekeLu: You are right, `c.cancel` is always called no matter whether `graceful` is true or false. Then we should move this check to line 247 as what you have done.

Good catch for the comment message. Lets put `chromedp.NewExecAllocator` first.

```
// c.cancel is nil when Cancel is wrongly called with a context returned 
// from chromedp.NewExecAllocator or chromedp.NewRemoteAllocator
```

ZekeLu: > Furthermore, this error no longer occurs if cancel is passed to the context in chromedp.NewRemoteAllocator or chromedp.NewExecAllocator.

It just hides the issue. If we do this, it will change the behavior of the `Cancel` method. If you think that we should change the behavior, please document the new behavior.

Valentin-Kaiser: I think you are right, I have reverted the changes.

chromedp.go
```diff
@@ -244,7 +244,9 @@ func FromContext(ctx context.Context) *Context {
 // underlying errors happening during cancellation.
 func Cancel(ctx context.Context) error {
 	c := FromContext(ctx)
-	if c == nil {
+	// c.cancel is nil when Cancel is wrongly called with a context returned
+	// by chromedp.NewExecAllocator or chromedp.NewRemoteAllocator.
+	if c == nil || c.cancel == nil {
```
ZekeLu, line 249: Now lets add a test to cover it. Add it after line 263: https://github.com/chromedp/chromedp/blob/97a00b98306467ee7efe556f09fbc9cac0e7c94c/chromedp_test.go#L260-L263

```go
	if err := Cancel(allocCtx); err != ErrInvalidContext {
		t.Fatalf("want error %q, got %q", ErrInvalidContext, err)
	}
```

ZekeLu (changes requested):

chromedp_test.go
```diff
@@ -261,6 +261,12 @@ func TestCancelError(t *testing.T) {
 		t.Fatalf("expected a nil error, got %v", err)
 	}
 
+	allocCtx, allocCancel := NewExecAllocator(context.Background(), allocOpts...)
+	defer allocCancel()
```
ZekeLu, line 265: I think we don't need a new `allocCtx`. Just use the one that already exists.

ZekeLu (changes requested):

chromedp_test.go
Valentin-Kaiser: You're right, I didn't realise it was already defined.

ZekeLu (approved):
LGTM. Thank you!

PR #1354, opened 2023-08-16
add ScreenshotScale that allows specifying page scale factor

### LINKED ISSUE #1191: How does chromdp scale html or svg files for screenshots
How does chromdp scale html or svg files for screenshots
If you can, could you give me an example? thank you

ZekeLu:
After the change, there will be four functions for taking screenshot:

```go
func Screenshot(sel interface{}, picbuf *[]byte, opts ...QueryOption) QueryAction
func ScreenshotScale(sel interface{}, scale float64, picbuf *[]byte, opts ...QueryOption) QueryAction
func CaptureScreenshot(res *[]byte) Action
func FullScreenshot(res *[]byte, quality int) EmulateAction
```

It's a little mess, I hope that we can come up with a better API to unify them all.

Fixes #1191.


mrueg:
Thanks a lot for creating this change! It will definitely be helpful for https://github.com/kovetskiy/mark :)

mrueg (approved):

Issue #1418, opened 2024-01-09
panic: sync: negative WaitGroup counter


skaslev:
We're using [gotenberg](https://github.com/gotenberg/gotenberg) to convert html to pdf and we're seeing occasional panics in chromedp such as

```
panic: sync: negative WaitGroup counter

goroutine 431735 [running]:
sync.(*WaitGroup).Add(0xc0001c0420?, 0xc000170a80?)
	/usr/local/go/src/sync/waitgroup.go:83 +0xda
sync.(*WaitGroup).Done(...)
	/usr/local/go/src/sync/waitgroup.go:108
github.com/chromedp/chromedp.(*ExecAllocator).Allocate.func2()
	/go/pkg/mod/github.com/chromedp/chromedp@v0.8.5/allocate.go:234 +0xd2
created by github.com/chromedp/chromedp.(*ExecAllocator).Allocate
	/go/pkg/mod/github.com/chromedp/chromedp@v0.8.5/allocate.go:216 +0xdf2
```

The problem might be caused by that the WaitGroup counter is increased twice [here](https://github.com/chromedp/chromedp/blob/b193297e042c86b7fe3cdcde2d91f39bf9a5a0f2/allocate.go#L213-L216) and then decreased by one [here](https://github.com/chromedp/chromedp/blob/b193297e042c86b7fe3cdcde2d91f39bf9a5a0f2/allocate.go#L235) and [here](https://github.com/chromedp/chromedp/blob/b193297e042c86b7fe3cdcde2d91f39bf9a5a0f2/allocate.go#L319).

Additionally, on error the WaitGroup counter is decreased again [here](https://github.com/chromedp/chromedp/blob/b193297e042c86b7fe3cdcde2d91f39bf9a5a0f2/allocate.go#L254) which might make the counter negative and cause a panic.

PR #1419, opened 2024-01-09
Fix 'panic: sync: negative WaitGroup counter'

skaslev:
Fixes issue #1418


skaslev:
@ZekeLu can you please take a look?

skaslev:
🙄

PR #1422, opened 2024-01-15
Bump deps and add a dependabot config

mrueg:
This bumps golang dependencies and adds a dependabot config that will update github workflows and go deps in the future.


mrueg:
@kenshaw if you have time for a quick review, would be great to get this in :)

PR #1466, opened 2024-04-12
Added page.EventDownloadProgress to ignored cases in target.go

coezbek:
Adding EventDownloadProgress to events which are ignored in the case enumeration to avoid getting spammed with errors while a download is progressing.

### DIFF
--- a/target.go
+++ b/target.go
@@ -246,6 +246,8 @@ func (t *Target) pageEvent(ev interface{}) {
 		return
 	case *page.EventDownloadWillBegin:
 		return
+	case *page.EventDownloadProgress:
+		return
 
 	default:
 		t.errf("unhandled page event %T", ev)


coezbek:
@kenshaw Was this PR misguided? It seemed like this events needs to be ignored to unhandled updates during downloads.

kenshaw:
I've closed out all the old PRs, simply based on age.

Separately:

No, your PR was not misguided. Most of the PRs in the last 2 years have just been about superfluous message output, which I would classify this as. This has actually always been something that could be disabled  with the `chromedp.WithErrorf` option which would allows the package user to ignore messages where needed. I had originally put in this logic when I first developed the package to ensure that `chromedp` was responding appropriately to the fundamental CDP events/messages and acting accordingly.

Since this is a continuing issue/problem (simply because the CDP continues to evolve) and because I do not want to have to continue adding "ignored" messages, I plan to overhaul this logic and to use Go's builtin structured logging.

PR #1509, opened 2024-09-26
Update to latest cdproto.

dragonator:


louisheath:
I'm also trying to bump this package so that in `domEvent` we can drop `EventScrollableFlagUpdated` instead of error logging it, which is noisy

istiak101:
This PR can be closed now.

PR #1548, opened 2025-03-05
fix: add page.EventFrameStartedNavigating to ignored events (Google Chrome 134.x)

gulien:
Hey 👋 

Since Google Chrome 134.x (amd64), [I get a lot of errors logs](https://github.com/gotenberg/gotenberg/issues/1139):

```
ERROR: unhandled page event *page.EventFrameStartedNavigating
```

I wonder if it should be handled somehow 🤔

### DIFF
--- a/target.go
+++ b/target.go
@@ -333,7 +333,8 @@ func (t *Target) pageEvent(ev any) {
 		*page.EventScreencastVisibilityChanged,
 		*page.EventWindowOpen,
 		*page.EventBackForwardCacheNotUsed,
-		*page.EventFrameSubtreeWillBeDetached:
+		*page.EventFrameSubtreeWillBeDetached,
+		*page.EventFrameStartedNavigating:
 		return
 
 	default:


kenshaw:
Merged and tagged this as `v0.13.1` -- it's worth noting, however I was not getting these messages on Google Chrome (Linux) v134. Are you using macOS, Windows, or some other platform?

gulien:
Thanks @kenshaw for the quick response! I confirm it now works as expected on `linux/amd64` when converting HTML to PDF 😄 

> Are you using macOS, Windows, or some other platform?

I was able to reproduce this error with an `linux/amd64` container, both on my demo environment (host = `linux/amd64`) and on my macOS laptop (using Docker flag `--platform=linux/amd64`).

It did not happen previously on Google Chome `133.x`, nor does it happen on Chromium `133.x` on other platforms (namely, `linux/arm64`, `linux/arm/v7`, `linux/i386`).

Thanks again!

PR #1596, opened 2025-09-25
chore: remove unused min/max func

spuradage:
Go 1.21 adds built-in functions min and max.   https://tip.golang.org/doc/go1.21
Therefore, we can directly remove our own implementations and use the built-in function with the same names.


spuradage:
@kenshaw Hi, Could you please review this PR at your convenience? Thank you very much.

caltechustc:
I made the same observation, but I saw that it has already been submitted.
Good fixes.  @spuradage 

CC @kenshaw

PR #1616, opened 2026-02-10
Fix AppArmor failure; make TestPollTimeout fail in time

karelbilek:
Currently ChromeDP test fail on CI on some AppArmor nonsense.

I have disabled AppArmor in the CI. Now the AppArmor nonsense doesn't stop ChromeDP anymore.

However, the tests still fail; some of them fail on my local machine too, so it doesn't all seem to be a CI issue anymore. It seems like "actual fails" mostly.

TestPollTimeout now times out forever, which blocks the CI for 10 minutes; I have capped it with `context.WithTimeout`. It still fails, though. (Doesn't fail on my local machine; it can be some timing thing.)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,6 +17,8 @@ jobs:
           sudo apt-get install -y build-essential
       - name: Checkout code
         uses: actions/checkout@v4
+      - name: Disable apparmor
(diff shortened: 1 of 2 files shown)


karelbilek:
using `no-sandbox` consistently everywhere seems to also fix the issue, but it's too much work; and it makes the examples too ugly, as they also need to explicitly set the `no-sandbox`.

PR #1631, opened 2026-06-08
Add HTTP header support for remote WebSocket dial

jiyeyuran:
## Summary

- Add `WithDialHTTPHeader` browser option to send custom HTTP headers on the WebSocket upgrade request when dialing a CDP endpoint.
- Add `WithRemoteDialHTTPHeader` remote allocator option that forwards headers to the browser dial path.
- Add tests verifying `Authorization` is sent (and absent by default).

## Motivation

Some hosted browser/CDP services (e.g. Cloudflare Browser Rendering) require authenticated WebSocket connections. Today chromedp dials remote debugger URLs without a way to attach headers such as `Authorization: Bearer <token>`, forcing callers to run a local WebSocket proxy.

## Usage

```go
hdr := http.Header{"Authorization": []string{"Bearer " + token}}
allocCtx, cancel := chromedp.NewRemoteAllocator(
    ctx, wsURL,
    chromedp.NoModifyURL,
    chromedp.WithRemoteDialHTTPHeader(hdr),
)
```

## Test plan

- [x] `go test -run TestDialHTTPHeader ./...`
- [x] `go vet ./...`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/allocate.go
+++ b/allocate.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


shynome:
I think support "set custom Client" is better

